PR #10, opened 2015-01-16
Auto add new packages.

ralphbean:
This should be the last missing piece we need before we can deploy
the-new-hotness.

It watches for new packages being added to Fedora (after they've gone
through review, when they get added to pkgdb), and then it checks to see if
any projects are registered in anitya with the same upstream_url.

If none exist, then it adds the new project with the best guess it can at
the backend and it maps the project to the new Fedora package.

We map the package if the project already exists.

Also, this publishes new fedmsg messages about success and failure.

The new fedmsg messages will require a new entry in the fedmsg.meta module to
handle them.

Lastly, this requires that a new release of release-monitoring.org get deployed
that includes the code from fedora-infra/anitya#74.

### DIFF
--- a/fedmsg.d/hotness-example.py
+++ b/fedmsg.d/hotness-example.py
@@ -48,6 +48,12 @@ config = {
         'target_tag': 'rawhide',
     },
 
+    'hotness.anitya': {
+        'url': 'https://release-monitoring.org',
+        #'username': '....',
+        #'password': '....',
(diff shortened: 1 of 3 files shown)


pypingou:
Couple of questions but won't change much, :+1:

ralphbean:
Thanks @pypingou!

PR #11, opened 2015-02-10
Suppress errors.

ralphbean:
Currently, we fail to kick off scratch builds somewhat often.
Sometimes:

- The spec file is formatted very strangely, and rpmdev-bumpspec doesn't
  know what to do.
- The SOURCES are not downloadable, and spectool doesn't know what to
  do.
- The modified file is not even buildable into an SRPM for koji, so
  rpmbuild doesn't know what to do.
- Koji is not reachable over the network because it's down or something,
  so the koji client api doesn't know what to do.

In all of these cases, we currently generate two emails:

- First, with a ``log.error()`` statement **inside**
  ``buildsys.handle(...)``.
- Second, with an exception that bubbles up to the fedmsg code above us.

This code catches the exception and just logs it as a **warning** since,
it's really not a problem with our code, but instead just with the
peculiarities of this or that Fedora package.

Best would be to catch these scenarios, and follow up in the ticket with
a statement like "Could not kick off a scratch build of the new release
for $REASON", but I'm just going to defer that for another day and go
with better server-side logging instead.

### DIFF
--- a/hotness/consumers.py
+++ b/hotness/consumers.py
@@ -20,6 +20,7 @@ Authors:    Ralph Bean <rbean@redhat.com>
 """
 
 import socket
+import traceback
 
 import fedmsg
 import fedmsg.consumers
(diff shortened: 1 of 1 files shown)


pypingou:
:+1:

PR #12, opened 2015-02-17
Only followup on bugs that are not already closed.

ralphbean:
See this one, where we posted about a rawhide build, but the bug was
already closed:  https://bugzilla.redhat.com/show_bug.cgi?id=969399#c67


pypingou:
Cool, :+1:

PR #15, opened 2015-02-20
Only followup on rawhide builds.

### LINKED ISSUE #14: Do not report official builds on bugzilla
Today ppisar was complaining on IRC about the number of emails sent by bugzilla but generated by the-new-hotness.

Before: https://bugzilla.redhat.com/1183577 
After: https://bugzilla.redhat.com/1194256

There are 5 new comments added by the-new-hotness, 4 of which are about the official builds.

While (unlike ppisar) I like the idea of having the results of the scratch-build sent to bugzilla, I think the other 4 comments about the official builds could be dropped.

Proposal:
Before commenting about a non-scratch-build check the comments for a previous message from us, if there is one, do nothing, if there isn't one, add one but then maybe only about rawhide and ignore the other branches.

ralphbean:
And we define what a rawhide build is in the config.

This will be, like so many other things, a PITA in that when mass branch
time comes along, we'll have to remember to go into the hotness config
and update the suffix to the new rawhide suffix.  Such is life (if we
had a releng microservice that made available what the rawhide tag and
suffix was.. that would be awesome.  We should write that).

This fixes #14.

### DIFF
--- a/fedmsg.d/hotness-example.py
+++ b/fedmsg.d/hotness-example.py
@@ -22,6 +22,11 @@ More information about the service that created this bug can be found at:
 config = {
     'hotness.bugzilla.enabled': True,
 
+    # Only followup about real builds with these release suffixes (rawhide)
+    'hotness.followup_suffixes': [
+        'fc23',
+    ],
+
     'hotness.bugzilla': {
         #'user': None,
         #'password': None,
--- a/fedmsg.d/hotness-example.py
+++ b/fedmsg.d/hotness-example.py
@@ -22,11 +22,6 @@ More information about the service that created this bug can be found at:
 config = {
     'hotness.bugzilla.enabled': True,
 
-    # Only followup about real builds with these release suffixes (rawhide)
-    'hotness.followup_suffixes': [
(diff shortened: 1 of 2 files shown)


pypingou:
There is pkgdb2 :-)

ralphbean:
> There is pkgdb2 :-)

Explain?

tyll:
On Fri, Feb 20, 2015 at 11:04:37AM -0800, Ralph Bean wrote:
> > There is pkgdb2 :-)
> 
> Explain?

pkgdb2 knows the disttag, so you can query it instead of setting it in
the code:
https://admin.fedoraproject.org/pkgdb/collection/master/

ralphbean:
Perfect.  https://admin.fedoraproject.org/pkgdb/api/collections/master/

Thank you.

ralphbean:
> you can query it instead of setting it in the code

This is done in cd3ce37.

pypingou:
:+1: for me

Note: we could invalidate the cache on receiving fedmsg message of an update of the collection in pkgdb.

PR #20, opened 2015-02-23
Use the kojira repos to get the latest rawhide info.

### LINKED ISSUE #16: Pull from different rawhide repo.
The hotness hears about a new upstream release being discovered, it compares the version of it against what it thinks is in rawhide.  It uses `repoquery` with [this yum config](https://infrastructure.fedoraproject.org/cgit/ansible.git/tree/roles/hotness/files/yumconfig) to figure out what we currently have.

The ``metadata_expire`` part is set intentionally small, so that shortly after something gets built for rawhide, hotness will know about it and not file bugs by mistake.  But, it's not working as expected. I'll release something mid-day and build it for rawhide, then 6 hours later anitya will run, notice the new upstream release, then hotness will run and not yet know about my new rawhide build.

To close this window, we can use the repo used internally by koji for buildroots (generated by kojira):

https://kojipkgs.fedoraproject.org/repos/rawhide/latest/

A new problem stemming from that, is that that repo doesn't have any SRPMs and the-new-hotness currently requires a source repo to do its check.  **So**, we simply need to modify the repoquery code to allow it to check the binary repos instead, then we can change our production config to use this new more frequently updated repo.

ralphbean:
This should fix #16.

I thought it was going to require modification to
``hotness.repository.build_nvr_dict``, but local tests indicate that no changes
are required.

### DIFF
--- a/yum-config
+++ b/yum-config
@@ -11,12 +11,26 @@ plugins=1
 installonly_limit=3
 
 # It is usually 90m.  We'll expire often..
-metadata_expire=10m
+metadata_expire=5m
 
-# We have only one repo to query...
-[rawhide-source]
-name=Rawhide Source
+# We have three repos to query...
+[rawhide-x86_64]
+name=Rawhide x86_64
 failovermethod=priority
-baseurl=http://dl.fedoraproject.org/pub/fedora/linux/development/rawhide/source/SRPMS/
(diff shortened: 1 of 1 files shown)


pypingou:
fwm, :+1:

PR #22, opened 2015-02-24
Generate a nicer changelog for the scratch task srpm.

### LINKED ISSUE #19: Changelog entry for scratch build is useless
It would be nice if the scratch build can contain some more useful entry. You can compare the-new-hotness changelog entry with previous entries for rubygem-jbuilder build for example [1].

Not mentioning that the formatting with quotes etc is suboptimal.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1154598#c6

ralphbean:
This should fix #19.

### DIFF
--- a/hotness/buildsys.py
+++ b/hotness/buildsys.py
@@ -96,7 +96,7 @@ class Koji(object):
             cmd = [
                 '/usr/bin/rpmdev-bumpspec',
                 '--new', upstream,
-                '-c', 'Update to %s (#%s)' % (upstream, rhbz.bug_id),
+                '-c', 'Update to %s (#%d)' % (upstream, rhbz.bug_id),
                 '-u', self.userstring,
                 specfile,
             ]
--- a/hotness/buildsys.py
+++ b/hotness/buildsys.py
@@ -96,8 +96,8 @@ class Koji(object):
             cmd = [
                 '/usr/bin/rpmdev-bumpspec',
                 '--new', upstream,
-                '-c', '"Latest upstream, %s for #%s"' % (upstream, rhbz),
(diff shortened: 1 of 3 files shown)


ralphbean:
@voxik, does this look like it will do it for you?

hotness/buildsys.py
```diff
@@ -96,8 +96,8 @@ def handle(self, package, upstream, version, rhbz):
             cmd = [
                 '/usr/bin/rpmdev-bumpspec',
                 '--new', upstream,
-                '-c', '"Latest upstream, %s for #%s"' % (upstream, rhbz),
-                '-u', '"%s"' % self.userstring,
+                '-c', 'Update to %s (#%s)' % (upstream, rhbz.bug_id),
```
relrod: Should this be `#%d` instead?

ralphbean: Why, yes it should, my type-oriented friend.

(fwiw, the ``%s`` works.. but ``%d`` would be more explicit).

relrod: > Why, yes it should, my type-oriented friend.

In a dependently-typed language (Idris, Agda, Coq, etc.) you could detect that mismatch at compile time ;). Not in Haskell though. :P

relrod: @ralphbean If you want to see how that works (it's really neat): https://www.youtube.com/watch?v=fVBck2Zngjo

voxik:
Looks good if I can parse the code correctly :+1:

pypingou:
:+1: for me as well

ralphbean:
Thanks both!

PR #24, opened 2015-02-24
For github backend, if the homepage is on github, specify the version_url to use

pypingou:
This version_url contains the owner/project information that is required by
anitya to automatically retrieve version information.


ralphbean:
LGTM!

PR #25, opened 2015-02-24
Suppress some errors from rpmbuild and friends.

ralphbean:
We get lots of errors every now and then from hotness, mostly from failures to
automatically bump the spec and build the srpm.  Most of these are things we
can't do anything about, so no need to email us every day about it.


pypingou:
Do we publish a message about these?

If so, :+1: otherwise might be something worth adding

ralphbean:
> Do we publish a message about these?

Not currently.

> otherwise might be something worth adding

Filed #26 for it.

PR #28, opened 2015-02-26
Strip leading v.

ralphbean:
Which is sometimes prefixed on github version tags.

Fixes fedora-infra/anitya#102


pypingou:
Nice, didn't know this one :+1:

PR #30, opened 2015-03-05
This should actually be cached.

ralphbean:
I think I removed this by mistake in 4f300982 when I was doing some
debug work.


pypingou:
:+1:

Issue #26, opened 2015-02-24
Publish a message when we fail to initiate a scratch build


ralphbean:
This could be because koji is down, or because we cannot build the srpm (because there are errors in the spec file).. lots of reasons.

Currently the errors just go to the local logs.  We should advertise it on fedmsg.

Issue #18, opened 2015-02-23
Scratch builds done just sometimes


voxik:
These are latest updates reported by new hotness for my packages:

https://bugzilla.redhat.com/show_bug.cgi?id=1115077
https://bugzilla.redhat.com/show_bug.cgi?id=1118694
https://bugzilla.redhat.com/show_bug.cgi?id=1154598

I am wondering, why the scratch build for rhbz#1118694 was not done.

pypingou:
Without having check the logs, could it be related to Source1?
http://pkgs.fedoraproject.org/cgit/rubygem-gettext_i18n_rails.git/tree/rubygem-gettext_i18n_rails.spec#n14

ralphbean:
> I am wondering, why the scratch build for rhbz#1118694 was not done.

Yeah, hotness failed to even build the SRPM before it could submit it to koji.  It currently doesn't follow up with any information if it fails that early.

Issue #5, opened 2014-11-21
Submit patch to bugzilla


ralphbean:
When hotness does its scratch build, it should also attach the patch that it generated to the bugzilla ticket that it created.

rahulsundaram:
Any update on this?

ralphbean:
None yet.  However, you can grab the srpm from the scratch build and use 'fedpkg import' (someone told me they were doing this.. I think it was @voxik).

rahulsundaram:
interesting!  thanks for the tip

PR #36, opened 2015-03-26
Create and attach patches to bz tickets we file.

ralphbean:
Also, propagate srpm-creation and koji-kickoff errors to the ticket, instead of
emailing sysadmin-datanommer.

Almost always, this is an error in the specfile and I'd like to make the
packager(s) aware of that since I can't really do anything about it myself.

### DIFF
--- a/hotness/buildsys.py
+++ b/hotness/buildsys.py
@@ -127,8 +127,8 @@ class Koji(object):
             filename = filename.strip()
 
             # Copy the patch out of this doomed dir so bz can find it
-            destination = '/var/tmp/' + filename
-            shutil.move('/'.join([tmp, filename]), destination)
+            destination = os.path.join('/var/tmp', filename)
+            shutil.move(os.path.join(tmp, filename), destination)
 
             return task_id, destination, '[patch] ' + comment
         finally:
--- a/hotness/buildsys.py
+++ b/hotness/buildsys.py
@@ -27,10 +27,6 @@ class Koji(object):
         self.opts = config['opts']
         self.priority = config['priority']
         self.target_tag = config['target_tag']
-        self.passable_errors = config.get('passable_errors', [
-            # This is the packager's problem, not ours.
-            'unclosed macro or bad line continuation',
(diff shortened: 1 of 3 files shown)


ralphbean:
Fixes #5.  Fixes #18.  Fixes #26.

hotness/buildsys.py
```diff
@@ -124,7 +118,19 @@ def handle(self, package, upstream, version, rhbz):
...
+            # Now, craft a patch to attach to the ticket
+            output = self.run(['git', 'commit', '-a',
+                               '-m', comment,
+                               '--author', self.userstring], cwd=tmp)
+            filename = self.run(['git', 'format-patch', 'HEAD^'], cwd=tmp)
+            filename = filename.strip()
+
+            # Copy the patch out of this doomed dir so bz can find it
+            destination = '/var/tmp/' + filename
+            shutil.move('/'.join([tmp, filename]), destination)
```
pypingou: This line scares me a little as it would accept a folder as well as a file, bu on the other side things are pretty contained about what they can do.

Btw, ``'/'.join()`` not ``os.path.join()``?

hotness/consumers.py
```diff
@@ -180,12 +180,20 @@ def handle_anitya(self, msg):
...
                 self.triggered_task_ids[task_id] = bz
-            except Exception:
-                self.log.warning("Failed to kick off scratch build.")
+
+                # Attach the patch to the ticket
+                self.bugzilla.attach_patch(patch_filename, description, bz)
+            except Exception as e:
+                heading = "Failed to kick off scratch build."
+                note = heading + "\n\n" + str(e)
+                self.log.warning(heading)
```
pypingou: Should it be ``warning(note)`` ?

pypingou:
One question, otherwise :+1:

hotness/consumers.py
ralphbean: Not necessarily.  The whole exception traceback gets logged on the very next line.

PR #38, opened 2015-03-27
Listen for pkgdb monitoring toggle events.

### LINKED ISSUE #32: Listen for pkgdb monitoring toggle messages
Problem:  Sometimes, a package owner adds a package to anitya, but the pkgdb monitoring flag is False.  Anitya finds a new version, broadcasts it, and the-new-hotness drops it, since the monitoring flag is false.  Shortly thereafter, the package owner turns the monitoring flag *on*, but there is nothing to trigger the-new-hotness to do anything *until* the next release upstream happens.

the-new-hotness could listen for pkgdb messages indicating that monitoring was turned on at which point it could go and actively query anitya to see if it should file any bugs, etc.

### LINKED ISSUE #35: Setting pkgdb monitor to true should trigger a new anitya addition
We currently watch for when new packages are added to Fedora (when they finish package review) and we add them to anitya at that point.

It would also be cool if when the monitor flag is set to True, we try to do the same thing.

ralphbean:
And use those events as another opportunity to try and map stuff in
anitya.

Fixes #32.
Fixes #35.

### DIFF
--- a/hotness/consumers.py
+++ b/hotness/consumers.py
@@ -305,18 +305,31 @@ class BugzillaTicketFiler(fedmsg.consumers.FedmsgConsumer):
             self.log.info("Found one match on Anitya.")
             project = projects[0]
             anitya.login(self.anitya_username, self.anitya_password)
-            success = anitya.map_new_package(name, project)
+            reason = None
+            try:
+                anitya.map_new_package(name, project)
+            except ValueError as e:
+                reason = str(e)
             self.publish("project.map", msg=dict(
-                trigger=msg, project=project, success=success))
+                trigger=msg,
+                project=project,
+                success=not bool(reason),
+                reason=reason))
             return
 
         # ... else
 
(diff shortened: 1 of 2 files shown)


hotness/consumers.py
```diff
@@ -311,6 +318,45 @@ def handle_new_package(self, msg):
...
+        if total > 1:
+            self.log.info("%i projects with %r %r already exist in anitya." % (
+                total, name, homepage))
+            return
+        elif total == 1:
+            # The project might exist in anitya, but it might not be mapped to
+            # a Fedora package yet, so map it if necessary.
+            project = results['projects'][0]
+
+            if not any([p['distro'] == 'Fedora' for p in project['packages']]):
```
pypingou: ``p['distro'].lower() == 'fedora'`` ?

pypingou:
:+1:

Issue #34, opened 2015-03-15
Listen for anitya added distro mapping messages


tyll:
If a new distro mapping is added to anitya which includes Fedora, the new hotness should check whether to report a new version to Fedora. See also #31

PR #39, opened 2015-03-28
Handle multiply mapped anitya projects.

### LINKED ISSUE #33: File multiple bugs if there are multiply mapped packages
See this case:

https://apps.fedoraproject.org/datagrepper/id?id=2015-fd27e2ea-2506-4516-9c10-8aab1c00c804&is_raw=true&size=extra-large

sqlalchemy is mapped to multiple packages in Fedora.  If so, then we should file a bug on each one **if** each one has the monitor flag set to True.  (we should also publish separate followup messages for each one).

ralphbean:
It is possible for an anitya project to be mapped to two or more Fedora
packages.  In these cases we will now check the monitoring flag of *each*
one of those packages (not just whichever one appears first in the
list) and then file bugs for each of those packages that **do** want
bugs to be filed.

Fixes #33.

### DIFF
--- a/hotness/anitya.py
+++ b/hotness/anitya.py
@@ -221,7 +221,7 @@ class Anitya(object):
         data = resp.json()
 
         if 'error' in data:
-            log.error('Anitya error: %r' % data['error'])
+            log.warning('Anitya error: %r' % data['error'])
         else:
             log.info("Check yielded upstream version %s for %s" % (
                 data['version'], data['name']))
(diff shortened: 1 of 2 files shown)


ralphbean:
4cef2de also fixes #34.

pypingou:
:+1: for me

ralphbean:
Thanks so much, @pypingou!

PR #45, opened 2015-05-20
Use the new is_monitored "nobuild" flag.

ralphbean:


hotness/consumers.py
```diff
@@ -213,6 +214,11 @@ def _handle_anitya_update(self, upstream, package, msg):
             self.publish("update.bug.file", msg=dict(
                 trigger=msg, bug=dict(bug_id=bz.bug_id)))
 
+            if is_monitored == 'nobuild':
+                self.log.info("Monitor flag set to 'nobuild'.  "
+                              "Skipping scratch build.")
```
pypingou: Do we want to ``self.publish("update.drop", msg=dict(trigger=msg, reason="pkgdb"))`` here as well?

pypingou:
One question, otherwise :+1:

hotness/consumers.py
ralphbean: Nah.  They have to turn this on explicitly -- so they should know already.

ralphbean:
Thanks @pypingou!

PR #47, opened 2015-06-11
mark patches as such when attaching them in Bugzilla

danc86:
When a Bugzilla attachment is flagged as a patch, Bugzilla
enables some extra UI features like diff display.


pypingou:
Oh cool :+1: for me but I'll let @ralphbean do the merge :)

ralphbean:
Nice!

Issue #56, opened 2015-08-27
Could not add anitya project. Not logged in.


pnemade:
I am seeing this kind of message for many other projects but could not understand why its coming besides some users saying they are already logged into Anitya.

https://apps.fedoraproject.org/datagrepper/raw?topic=org.fedoraproject.prod.hotness.project.map

Can someone help to understand this?

tomhughes:
I just got this for libosmium. At 1214 UTC I got this:

```
A new version of "libosmium" has been detected:  "v2.4.0" newer than "v2.3.0", packaged as "libosmium"
	https://release-monitoring.org/project/7023/
```

but no bug was filed despite "bugs and builds" being set at https://admin.fedoraproject.org/pkgdb/package/libosmium/ so I tried turned that off and on again to force a new check and then at 1518 UTC I got this:

```
hotness tried to map libosmium to an upstream project, but failed:  "Could not add anitya project.  Not logged in."
	https://release-monitoring.org/project/7023/
```

pnemade:
Please someone from developers reply. I think this is happening since last 3 weeks now.
@ralphbean can you help here?

ralphbean:
Sorry, been busy with bodhi2 here.  I think it has a session with anitya that just expires.  Some logic will need to be added that tells it how to log back in when that happens.  Let me restart it and see if that fixes it (short term).

ralphbean:
Well, I tried that and then tried toggling the monitor flag on libosmium again, but it didn't work:  https://apps.fedoraproject.org/datagrepper/raw?package=libosmium

hotness still thinks that it isn't logged in, so, something else must be up.

ralphbean:
Can duplicate the login failure using the ``hotness.anitya`` module on my laptop.

ralphbean:
Using this to test, if anyone else wants to mess with it in the meantime:

```python
import hotness.anitya

client = hotness.anitya.Anitya()
client.login("YOUR_USERNAME", "YOUR_PASSWORD")
print client.is_logged_in
```

ralphbean:
@puiterwijk, @pypingou, do either of you have any ideas on this one?  I'm stuck for the moment.

tomhughes:
It seems that when `login` follows the redirect back from id.fedoraproject.org to release-monitoring.org the arguments it passes are not enough to convince release-monitoring.org that we are no logged in.

puiterwijk:
I have patches for both sides of the story, sending them in

[...]

PR #57, opened 2015-09-01
Use the root url to check for logged in state

puiterwijk:
The /login/fedora link may always just redirect the user silently.

Helps with #56, other part of the fix is in anitya.

Signed-off-by: Patrick Uiterwijk <puiterwijk@redhat.com>


pypingou:
:+1: for me (I had done something similar locally when trying to debug the issue)

Issue #65, opened 2015-09-22
If the upstream_url from pkgdb is empty, tnh should bail out early


ralphbean:
Relates to #46, in addition to whatever we come up with here.

tnh can't do anything intelligent about mapping the project in anitya if it has no upstream_url to work with.

PR #66, opened 2015-09-22
Handle pkgdb.package.update messages.

### LINKED ISSUE #46: hotness tried to map rubygem-ejs to an upstream project, but failed due to ambiguity.
I received following FMN notification when requested rubygem-ejs in pkgdb:

```
hotness tried to map rubygem-ejs to an upstream project, but failed due to ambiguity.  6436 other projects share the same homepage
```

I am not really sure what does it complain about ...

ralphbean:
This should fix #46 and #65.

Sometimes, a new package added to Fedora pkgdb doesn't have an upstream_url
off the bat.  We already guard against that and avoid trying to update
anitya with that empty information.

This new code responds when the pkgdb weekly cronjob goes through and
updates the upstream_url field from a scan of the .spec/srpm information.

Since packagers have indicated that a package has a new upstream url, we
use that to update anitya's homepage information, and potentially map
packages to projects for the first time.

Also, silence some BeautifulSoup warnings.

### DIFF
--- a/hotness/anitya.py
+++ b/hotness/anitya.py
@@ -214,11 +214,15 @@ class Anitya(object):
         response = self.__send_request(url, method='GET')
         return response.json()
 
-    def search_by_package(self, name):
-        url = '{0}/api/projects/?pattern={1}'.format(self.url, name)
+    def get_project_by_package(self, name):
+        url = '{0}/api/project/Fedora/{1}'.format(self.url, name)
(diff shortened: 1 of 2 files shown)


ralphbean:
@pypingou, OK.  pushed some commits that should address all your comments.

pypingou:
:+1: for me

ralphbean:
Thanks!

Issue #43, opened 2015-04-17
Failed to kick off scratch build.


jasontibbitts:
Was looking through some bugzilla tickets and I noticed the following error:

Failed to kick off scratch build.
list index out of range

in https://bugzilla.redhat.com/show_bug.cgi?id=1180445#c6
Not sure what caused it or if it's repeatable.

jasontibbitts:
Just saw this again, in https://bugzilla.redhat.com/show_bug.cgi?id=1217014

ralphbean:
I'm not sure what's going on here.  The 'list index out of range' is not enough information to go off of.  We need to:

- grep the logs for that error and try and find a traceback to get more context.

If we can't find that error in the logs, we'll have to close it as wontfix until we see it again (assuming that it has been fixed by some other commit in the meantime).

ralphbean:
I looked around, but there weren't any list index out of range errors in the logs that could be related to this.  I'll close as wontfix, hoping that it was fixed inadvertently a while ago.  If it happens again, please re-open.

PR #68, opened 2015-09-23
Be extra careful with anitya error panels.

ralphbean:
The only 'list index out of range' errors I see in the logs all come
from these code blocks.  I went looking for them while trying to fix
issue #43, but these code paths aren't related to that issue.


pypingou:
Sweet, :+1:

PR #70, opened 2015-09-24
Rename this to match the function definition.

ralphbean:
We renamed the definition of this and we renamed one other usage, but we forgot this spot.  Getting traceback emails about it.


pypingou:
:+1: easyfix

PR #75, opened 2015-10-01
Don't act on packages that are retired.

### LINKED ISSUE #72: Don't act on packages if they are retired in rawhide
We have code that tries to find the latest version in rawhide, to compare it with the upstream version found by anitya.

However, if the package is retired in rawhide, that version-finding code path will fail, and email us (annoying).

This happened today with the 'ekg' package:  https://apps.fedoraproject.org/datagrepper/raw?package=ekg&not_category=anitya

We should just be able to modify the `is_monitored` function to check for retirement status in addition to the monitor flag.

ralphbean:
This should fix #72.

### DIFF
--- a/hotness/consumers.py
+++ b/hotness/consumers.py
@@ -548,6 +548,13 @@ class BugzillaTicketFiler(fedmsg.consumers.FedmsgConsumer):
         try:
             data = r.json()
             package = data['packages'][0]
+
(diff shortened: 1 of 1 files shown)


pypingou:
Big :+1: I wanted to check for this myself, thanks for doing it :)

ralphbean:
woot!  :shipit:

PR #76, opened 2015-10-01
Silence this error email.

ralphbean:
And replace it with a warning in the logs.

I looked into all the error emails we got about this the last week, and
they're all reasonable situations.

- A new package was added to pkgdb after the Review process completed.
- hotness noticed that, and added the pkg to anitya.
- hotness forced an anitya check
- anitya published a fedmsg message that a new version was found
- hotness received that message and tried to compare that new upstream
  version against the version in rawhide but... of course, no build had
  completed in rawhide yet.  And even if it had in the tiny interval of
  time that has elapsed so far, the rawhide repo would not have been
  rebuilt for us to check.

So, we should assume that rawhide *will* be up to date with the latest
upstream at this point, since the package **just** came out of package
review.

### DIFF
--- a/hotness/consumers.py
+++ b/hotness/consumers.py
@@ -204,7 +204,19 @@ class BugzillaTicketFiler(fedmsg.consumers.FedmsgConsumer):
 
         # Is it new to us?
         fname = self.yumconfig
-        version, release = hotness.repository.get_version(package, fname)
+        try:
+            version, release = hotness.repository.get_version(package, fname)
+        except KeyError:
+            # At this point, we have tried very hard to find the rawhide
+            # version of the package.  If we didn't find it, that means there
+            # likely hasn't yet been a rawhide build of it.. and there's
+            # nothing reasonable we can do.  Notify the world of our failure
+            # and go back to the event loop.
+            self.log.warn("No rawhide version found for %r" % package)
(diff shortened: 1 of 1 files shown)


pypingou:
:+1: for me

hotness/consumers.py
```diff
@@ -204,7 +204,19 @@ def _handle_anitya_update(self, upstream, package, msg):
...
+        try:
+            version, release = hotness.repository.get_version(package, fname)
+        except KeyError:
+            # At this point, we have tried very hard to find the rawhide
+            # version of the package.  If we didn't find it, that means there
+            # likely hasn't yet been a rawhide build of it.. and there's
+            # nothing reasonable we can do.  Notify the world of our failure
+            # and go back to the event loop.
+            self.log.warn("No rawhide version found for %r" % package)
+            self.publish("update.drop", msg=dict(
```
pypingou: Does this need adjustment in fedmsg-meta?

ralphbean: Surprisingly, no.

Only a little [right here](https://github.com/fedora-infra/fedmsg_meta_fedora_infrastructure/blob/develop/fedmsg_meta_fedora_infrastructure/hotness.py#L103-L111).

ralphbean:
Thanks @pypingou!

PR #79, opened 2015-10-08
replace YUM with DNF

phracek:

This pull request replaces YUM with DNF.

### DIFF
--- a/hotness/buildsys.py
+++ b/hotness/buildsys.py
@@ -12,6 +12,8 @@ import time
 import koji
 import sh
 
+from rebasehelper.upstream_monitoring import UpstreamMonitoring
+
 
 class Koji(object):
     def __init__(self, consumer, config):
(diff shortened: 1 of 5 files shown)


ralphbean:
We need to be able to handle both.  :(

The node we run this on in fedora infra is a RHEL7 box.

Can you make it configurable?

Have you given it a running test with the dnf tools?

pypingou:
@phracek need help with this one?

phracek:
Sure I will update my PR. Both should be possible.

pypingou:
Hm, looks like this is changing more than what the title says. Shouldn't we keep those two changes separate and in two different PRs ?

phracek:
Sure but how to do it. It was done automatically via web:(

Of course two PR are the best.

pypingou:
Ah, you might have to use git for this, just keep the two set of changes into two different branches.

The current branch you're using for the yum/dnf change and another branch for the rebasehelper

phracek:
ok. will do it.

phracek:
By the way, shall I revert "rebase-helper integration" commit?

How to handle with it before I will introduce 'rebase-helper' branch.

phracek:
I think that easiest way is to close PR.

PR #81, opened 2015-10-15
When sending a comment to bugzilla add a link to the project in anitya

pypingou:
Fixes https://github.com/fedora-infra/anitya/issues/149


ralphbean:
LGTM.  We'll have to update the hotness config in ansible as well.  (hotness is not frozen, so we can do that asap).

pypingou:
Indeed but only once we did a release, merging for now, thanks for the review :)

PR #86, opened 2015-11-23
Fix release-monitoring.org html scraping.

ralphbean:
Kinda ugly here, but login with release-monitoring.org broke when we
added a search form to every page there.

Hotness would try to scrape each page for a form, and then extract the
csrf_token from that form.  Except now, there are **two** forms on every
page, so that scraping broke.

This code updates each attempt at that to look explicitly for the form
it is interested in.

### DIFF
--- a/hotness/anitya.py
+++ b/hotness/anitya.py
@@ -239,7 +239,8 @@ class Anitya(object):
 
         data = copy.copy(project)
         data['homepage'] = homepage
-        data['csrf_token'] = soup.form.find(id='csrf_token').attrs['value']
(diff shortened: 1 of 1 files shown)


pypingou:
Stupid idea: can we look for the ``csrf_token`` field anywhere in the page (ie: without searching for a specific form first)? Even if there are multiple ``csrf_token`` on the page, any of them should work.

ralphbean:
> Stupid idea

Good idea!  Done in 54c132d.

pypingou:
Cool, :+1:

PR #87, opened 2015-11-30
PR  which supports YUM and DNF usage

phracek:
This Pull Request supports as YUM as DNF.

It includes also SPEC and setup.py modifications.

### DIFF
--- a/fedmsg.d/hotness-example.py
+++ b/fedmsg.d/hotness-example.py
@@ -1,6 +1,36 @@
 import os
 import socket
+import ConfigParser
+from six import StringIO
+
+
 hostname = socket.gethostname().split('.', 1)[0]
+
+
+thn_section = 'thn'
+
+
+class ThnConfigParser(ConfigParser.ConfigParser):
+    def read(self, filename):
+        try:
+            text = open(filename).read()
+        except IOError:
+            pass
+        else:
+            section = "[%s]\n" % thn_section
+            file = StringIO(section + text)
+            self.readfp(file, filename)
+
+
+def get_pkg_manager():
+    release_file = '/etc/os-release'
+    config = ThnConfigParser()
+    config.read(release_file)
+    name = config.get(thn_section, 'ID')
+    if name == 'fedora':
+        return ["/usr/bin/dnf", "repoquery"]
+    else:
+        return ["/usr/bin/repoquery"]
 
 description_template = """Latest upstream release: %(latest_upstream)s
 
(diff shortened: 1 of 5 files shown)


ralphbean:
Cool.  So, this expects to find a new section in ``/etc/os-release`` specifically for the-new-hotness?

.. and if that section is not present, it defaults to ``yum``, correct?

ralphbean:
Hm, would it be easier to just check for the existence of ``/usr/bin/dnf`` and use that if it is present, otherwise use ``/usr/bin/yum``?

pypingou:
I was wondering about just providing the command to run in the configuration file, something like:
``cmd='/usr/bin/repoquery'`` vs ``cmd=/usr/bin/dnf repoquery`` since the rest of the command is the same.

phracek:
@pypingou ```cmd='/usr/bin/repoquery'``` doesn't work with DNF.

But it could be simplified. Sure.

And of course check for ```/usr/bin/dnf``` is much simpler then parsing /etc/os-release. EPEL7 uses YUM and F22-> uses dnf.

pypingou:
> @pypingou cmd='/usr/bin/repoquery' doesn't work with DNF.

I agree, but if you set the command in the configuration file it's easy to switch a variable to another:
from ``cmd = '/usr/bin/repoquery'`` to ``cmd = '/usr/bin/dnf repoquer``

phracek:
You are right.

phracek:
What configuration file shall I use? I don't see it.

ralphbean:
How about [this one](https://github.com/fedora-infra/the-new-hotness/blob/develop/fedmsg.d/hotness-example.py)?

ralphbean:
It gets loaded as ``self.config`` [here](https://github.com/fedora-infra/the-new-hotness/blob/develop/hotness/consumers.py#L81).

phracek:
Ok, I will update PR. Thanks

phracek:
Unfortunately on my F23 I see following:

~~~
unused-4-101: ~$ ls -la /usr/bin/{yum,dnf}
lrwxrwxrwx. 1 root root    5 Nov 27 11:25 /usr/bin/dnf -> dnf-3
-rwxr-xr-x. 1 root root 1447 Nov 16 15:59 /usr/bin/yum
unused-4-101: ~$ 
~~~

Therefore I guess that detection of binary is not enough.

phracek:
But the whole functionality, like use dnf or yum, can be moved to config.

pypingou:
> But the whole functionality, like use dnf or yum, can be moved to config.

Might be just as easy :)

ralphbean:
Any luck moving this to the [config](https://github.com/fedora-infra/the-new-hotness/blob/develop/fedmsg.d/hotness-example.py)?  Values added there become available as ``self.config`` [here](https://github.com/fedora-infra/the-new-hotness/blob/develop/hotness/consumers.py#L81).

phracek:
Sorry guys, I was a pretty busy. I plan to do it Wednesday or Thursday.

ralphbean:
No rush -- just trying to keep communication channels open.  :)

ralphbean:
I think this is handled elsewhere now (in the rebase-helper PR).

PR #94, opened 2016-01-13
Rebase helper integration.

phracek:
I have finally implemented.
https://partner-bugzilla.redhat.com/show_bug.cgi?id=1259391 Especially from 13.1.2016

Feel free to send comments and I will correct it.
If something is wrong do not hesitate to contact me.

### DIFF
--- a/hotness/buildsys.py
+++ b/hotness/buildsys.py
@@ -173,6 +173,7 @@ class Koji(object):
         self.log.info("Rebase-helper is going to rebase package")
         rh_stuff = {}
         result_rh = -1
+        rh_app = None
         try:
             url = self.git_url.format(package=package)
             self.log.info("Cloning %r to %r" % (url, tmp))
@@ -186,11 +187,17 @@ class Koji(object):
             result_rh = rh_app.run()
             rh_stuff = rh_app.get_rebasehelper_data()
             self.log.info("Rebasehelper finish properly")
-            self.log.info(rh_stuff)
 
         except Exception as ex:
             self.log.info('Rebase helper failed with unknown reason. %s' % ex)
-            return result_rh, rh_stuff
+            rh_stuff = rh_app.get_rebasehelper_data()
 
         return result_rh, rh_stuff
 
+    def rebase_helper_checkers(self, task_ids, tmp_dir):
+        argument = ['--non-interactive']
+        cli = CLI(argument)
+        rh_app = Application(cli)
+        rh_stuff = rh_app.run_download_compare(task_ids, tmp_dir)
+
+        return rh_stuff
--- a/hotness/buildsys.py
+++ b/hotness/buildsys.py
@@ -174,30 +174,34 @@ class Koji(object):
         rh_stuff = {}
         result_rh = -1
         rh_app = None
+        url = self.git_url.format(package=package)
+        self.log.info("Cloning %r to %r" % (url, tmp))
+        sh.git.clone(url, tmp)
+        os.chdir(tmp)
         try:
-            url = self.git_url.format(package=package)
-            self.log.info("Cloning %r to %r" % (url, tmp))
-            sh.git.clone(url, tmp)
-            os.chdir(tmp)
-            self.log.info("Rebasehelper package %s %s" % (package, upstream))
             argument = ['--non-interactive', '--buildtool', 'fedpkg', upstream]
             cli = CLI(argument)
             rh_app = Application(cli)
             rh_app.set_upstream_monitoring()
+            self.log.info("Rebasehelper package %s %s" % (package, upstream))
             result_rh = rh_app.run()
-            rh_stuff = rh_app.get_rebasehelper_data()
             self.log.info("Rebasehelper finish properly")
 
         except Exception as ex:
-            self.log.info('Rebase helper failed with unknown reason. %s' % ex)
-            rh_stuff = rh_app.get_rebasehelper_data()
-
+            self.log.info('builsys.py: Rebase helper failed with unknown reason. %s' % str(ex))
+        rh_stuff = rh_app.get_rebasehelper_data()
         return result_rh, rh_stuff
 
     def rebase_helper_checkers(self, task_ids, tmp_dir):
-        argument = ['--non-interactive']
-        cli = CLI(argument)
-        rh_app = Application(cli)
-        rh_stuff = rh_app.run_download_compare(task_ids, tmp_dir)
+        argument = ['--non-interactive', task_ids['new'][1]]
+        rh_app = None
+        rh_stuff = None
+        try:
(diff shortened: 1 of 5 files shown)


phracek:
I don't know why Support YUM and DNF is here. It seems that I did something wrong.

hotness/consumers.py
```diff
@@ -243,23 +252,68 @@ def _handle_anitya_update(self, upstream, package, msg):
...
-                note = heading + "\n\n" + str(e)
-                self.log.warning(heading)
-                self.log.warning(traceback.format_exc())
-                self.bugzilla.follow_up(note, bz)
+                tmp = tempfile.mkdtemp(prefix='thn-rh', dir='/var/tmp')
+                result_rh, rh_stuff = self.buildsys.rebase_helper(package, upstream, tmp, bz)
+                if int(result_rh) == 0:
+                    self.log.info('Rebase package %s to %s was SUCCESSFULL' % (package, version))
+                    for number, reference in six.iteritems(rh_stuff['build_logs'],):
+                        note = 'Scratch build completed %s' % reference
```
ralphbean: It looks like ``reference`` is a list, so that this ends up being printed in the ticket as ``['http://koji.fedoraproject.....']``.  Can you iterate over the list of references so that they are printed cleanly in the ticket?  Preferably each on a new line in the improbable case that there is more than one build log.

hotness/consumers.py
```diff
@@ -243,23 +252,68 @@ def _handle_anitya_update(self, upstream, package, msg):
...
+                        self.bugzilla.follow_up(note, bz)
+                else:
+                    self.log.info('Rebase package %s to %s FAILED. See for details' % (package, version))
+                    self.bugzilla.follow_up('Rebase package %s to %s FAILED' % (package, version), bz)
+                    for number, reference in six.iteritems(rh_stuff['build_logs']):
```
ralphbean: If a scratch build fails, it is possible that the reason is buried in ``root.log`` instead of ``build.log``.  Can you link to all logs *as well as* linking to the scratch build itself.  That was users can pick and choose which link they click to get to the info they need.

hotness/buildsys.py
```diff
@@ -152,9 +157,40 @@ def handle(self, package, upstream, version, rhbz):
...
+            url = self.git_url.format(package=package)
+            self.log.info("Cloning %r to %r" % (url, tmp))
+            sh.git.clone(url, tmp)
+            os.chdir(tmp)
+            self.log.info("Rebasehelper package %s %s" % (package, upstream))
+            argument = ['--non-interactive', '--buildtool', 'fedpkg', upstream]
+            cli = CLI(argument)
+            rh_app = Application(cli)
+            rh_app.set_upstream_monitoring()
+            result_rh = rh_app.run()
```
ralphbean: Does this call to ``.run()`` create a koji scratch task and then wait for it to complete?

If so, this could be problematic for the-new-hotness running in production.

The way it worked before was that:

- hotness would receive a message from anitya, try to bump the spec, kick off a scratch build, remember the task id, and then it **returned and went back to its main loop**.
- while koji was working on that scratch task, the-new-hotness would continue processing *other* messages... potentially creating more scratch tasks in the mean time.
- when koji finally finished that original scratch task, it would publish a message, hotness would notice, and then follow up on the bug it filed.

If rebase-helper waits for koji to finish each task, I don't think the-new-hotness will drown in its workload.  It will sit there waiting for koji to finish a task, meanwhile thousands of messages will pile up in its inbox.

hotness/consumers.py
```diff
@@ -297,8 +351,8 @@ def handle_buildsys_scratch(self, msg):
         text2 = "%s %s" % (subtitle, url)
 
         # Followup on bugs we filed
-        if task_id in self.triggered_task_ids:
-            bugs.append((self.triggered_task_ids.pop(task_id), text1))
+        #if task_id in self.triggered_task_ids:
+        #    bugs.append((self.triggered_task_ids.pop(task_id), text1))
```
ralphbean: Yeah, right here is where hotness would pick back up on a task that it kicked off previously.

hotness/buildsys.py
phracek: Yes, .run() create a koji scratch task and then wait for it to complete.
I will look on it from developer point of view and rewrite it a bit.

hotness/consumers.py
phracek: Well, 
I can return from rebase-helper task_id but for this case
I need to download all rpm packages for comparing packages by rpmdiff, abipkgdiff, pkgdiff, etc.

Is it possible to do it?
Or better say how can I download packages which are done by fedpkg.
I will call rebase-helper function for comparing.

ralphbean: I think we only used fedpkg to pull down the tarball and any patches.  We would interact with the koji xmlrpc api to kick off the build -- and it's possible (if a little tricky) to use that koji api to pull down all the builds artifacts at the end.  I expect that this is what rebase-helper is already doing.

ralphbean: Can the rebase helper logic be split into two parts?  One part that prepares the scratch build, submits it, and returns the task_id, and then another part that downloads the results and analyzes them?

phracek: yes, sure. It takes a time. But definitely yes. I go ahead.

phracek:
Rebase-helper functionality is not split yet, though;)

phracek:
I have a question regarding this. Rebase-helper now returns a two koji task_ids. First one for old sources and second one for new sources.

Le't say that I will add these to task_ids into self.triggered_task_ids. But when both builds finishes
how can I recognize that they belongs to rebase-helper builds?

ralphbean:
Just FYI, @phracek and I talked it out in IRC and have a plan.

phracek:
Scratch builds are now separated.
Rebase-helper only starts a RPM scratch builds and returns task_ids.
I have to tested behavior when scratch builds are finished and what's going on.

phracek:
It seems that fedmsg-hub does not get an info that scratch build finished.
I will need a deep analyse.

phracek:
The last code is not so Python cool, I will do it later on.

phracek:
Finally I have finish implementation. It seem to be working. It was great task to split rebase-helper into two parts.

Please review it and let me know what should be changed.
In the meantime I am going to release a new version of rebase-helper including EPEL-7 package so that you are able to integrate it into the-new-hotness process machine..

phracek:
Hi folks, Finally I have built rebase-helper for EPEL-7. 
Do you have a time frame when the PR will be closed?

https://apps.fedoraproject.org/packages/rebase-helper/
http://koji.fedoraproject.org/koji/buildinfo?buildID=737683

ralphbean:
Let's go for it.  :)

PR #103, opened 2016-02-23
Make this log statement less verbose.

ralphbean:
Any calls to ``log.error`` or ``log.exception`` will generate an email
to the admins of the-new-hotness.  Those are very useful, but if we get
them for non-urgent situations, it can inadvertently train us to not pay
attention to the emails over time.

So, I'd like to propose that we *demote* this log statement to a
warning.  What do you think @phracek?  Should it be kept as an error?

### DIFF
--- a/hotness/buildsys.py
+++ b/hotness/buildsys.py
@@ -210,8 +210,8 @@ class Koji(object):
             if int(ret_code) != 0:
                 self.log.warn('Comparing package were not successful')
             rh_stuff = rh_app.get_rebasehelper_data()
-        except Exception as ex:
-            self.log.info('Compare packages failed. Reason %s' % str(ex))
+        except Exception:
+            self.log.exception('Compare packages failed.')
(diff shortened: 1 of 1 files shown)


ralphbean:
In dbe3d62, I made a further change to complain more loudly if we get an exception from rebase-helper.  *That* log statement will generate an email to the admins (me) and it will automatically include the traceback of the exception, so we can debug it if there is a problem.

Medium term, I'd like to add @phracek and @thozza to the email list for those errors, if you wouldn't mind.

thozza:
I'm fine with the email @ralphbean.

phracek:
@ralphbean Definitely.. You can change to warning. Error is not correct. I agree with your commits.
And the mail would be fine.

ralphbean:
Cool!  Let's do it.  I'll try to have the emails coming your way today.  If not today, then Friday.  Thank you!

PR #104, opened 2016-02-29
Handle repoquery differently for yum and dnf.

ralphbean:
Hit this in staging.

I didn't hit the failure on my box because I was using dnf.


pypingou:
:+1: cool for me

PR #108, opened 2016-03-09
Fixes #107: Detect if file exists or is not empty

### LINKED ISSUE #107: Attach only files which exist or are not empty
I guess, the-new-hotness should be able to detect if attached file exists or even if it is not empty.
https://github.com/fedora-infra/the-new-hotness/blob/develop/hotness/bz.py#L125

I will send a PR.
This is relevant not only from rebase-helper side but generally, I guess.

phracek:

The fix should detect if file exists and is not empty.
This is not relevant only for rebase-helper case but general, I guess.

But of course, rebase-helper is going to be updated so that it provides logs which exists and are not empty.

### DIFF
--- a/hotness/bz.py
+++ b/hotness/bz.py
@@ -23,7 +23,7 @@ import copy
 import logging
 
 import bugzilla
-
+import os
 
 class Bugzilla(object):
     base_query = {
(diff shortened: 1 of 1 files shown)


ralphbean:
Looks good here!  :+1: to merge.

PR #106, opened 2016-03-07
Fixes #105: Rebase helper logs

phracek:

PR #109, opened 2016-03-09
Correct another instance of mis-used six.iteritems().

ralphbean:
I'd like @phracek to review this one before it gets merged since it
depends on the structure of the output from rebase-helper.  We found an
erroneous instance like this in #106 and I'm not sure if this is the
same kind of situation or not.

See https://github.com/fedora-infra/the-new-hotness/pull/106#discussion_r55363099


phracek:
Great, thanks. Another my mistake. Please merge it.

ralphbean:
Thanks @phracek!

PR #111, opened 2016-03-11
Fixes #110: This does not really fix the problem. Log about attaching is

### LINKED ISSUE #110: Bug 1316528 does not have scratch build attached/linked and See logs and attachments in this bugzilla global name 'os' is not defined
The bug https://bugzilla.redhat.com/show_bug.cgi?id=1316528 was filed for mod_auth_openidc.

In https://bugzilla.redhat.com/show_bug.cgi?id=1316528#c1 it says

```
Patching or scratch build for mod_auth_openidc and version 1.8.7 FAILED.
See for details
```

but it's not clear what should be seen.

In https://bugzilla.redhat.com/show_bug.cgi?id=1316528#c2 it says

```
Rebase helper failed.
See logs and attachments in this bugzilla global name 'os' is not defined
```

which seems like the-new-hotness (or something related) failed, not the builds.

If this is not the correct project to report this sort of errors, please amend https://fedoraproject.org/wiki/Upstream_release_monitoring with some links about the bugzillas being filed.

phracek:
valid only in case really attach.

### DIFF
--- a/hotness/bz.py
+++ b/hotness/bz.py
@@ -123,10 +123,10 @@ class Bugzilla(object):
         self.log.info("Followed up on bug: %s" % bug.weburl)
 
     def attach_patch(self, filename, description, bug):
-        self.log.debug("Attaching patch to bug %r" % bug.bug_id)
         if os.path.exists(filename) and os.path.getsize(filename) != 0:
+            self.log.debug("Attaching patch to bug %r" % bug.bug_id)
             self.bugzilla.attachfile(bug.bug_id, filename, description,
-                is_patch=True)
+                    is_patch=True)
             self.log.info("Attached patch to bug: %s" % bug.weburl)
 
     def ftbfs_bugs(self, name):


hotness/bz.py
```diff
@@ -123,10 +123,10 @@ def follow_up(self, text, bug):
         self.log.info("Followed up on bug: %s" % bug.weburl)
 
     def attach_patch(self, filename, description, bug):
-        self.log.debug("Attaching patch to bug %r" % bug.bug_id)
         if os.path.exists(filename) and os.path.getsize(filename) != 0:
+            self.log.debug("Attaching patch to bug %r" % bug.bug_id)
             self.bugzilla.attachfile(bug.bug_id, filename, description,
-                is_patch=True)
+                    is_patch=True)
```
ralphbean: Just a thought.  Since this isn't really a patch (but is instead a log file), should we pass ``is_patch=False``?  I don't know what that flag does on the bugzilla.redhat.com side.

ralphbean:
:+1:

PR #114, opened 2016-03-15
Fix #113 Text in bugzilla has to be clear.

### LINKED ISSUE #113: Bug 1317494 says "See for details"
This might or might not be related to https://github.com/fedora-infra/the-new-hotness/issues/110.

In https://bugzilla.redhat.com/show_bug.cgi?id=1317494#c1, upstream-release-monitoring@fedoraproject.org says

```
Patching or scratch build for perl-DBD-XBase and version 1.05 FAILED.
See for details
```

It is not clear what should be seen and where.

phracek:

### DIFF
--- a/hotness/consumers.py
+++ b/hotness/consumers.py
@@ -284,8 +284,7 @@ class BugzillaTicketFiler(fedmsg.consumers.FedmsgConsumer):
                                 else:
                                     self.new_triggered_task_ids[task_id] = [bz, None, str(upstream), str(package)]
                 else:
-                    note = 'Patching or scratch build for %s and version %s FAILED.\n' \
-                           'See for details' % (package, version)
+                    note = 'Patching or scratch build for %s and version %s FAILED.\n' % (package, version)
                     self.bugzilla.follow_up(note, bz)
                     if 'logs' in rh_stuff['build_logs']:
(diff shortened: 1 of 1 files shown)


ralphbean:
:+1: here

hotness/consumers.py
```diff
@@ -284,8 +284,7 @@ def _handle_anitya_update(self, upstream, package, msg):
                                 else:
                                     self.new_triggered_task_ids[task_id] = [bz, None, str(upstream), str(package)]
                 else:
-                    note = 'Patching or scratch build for %s and version %s FAILED.\n' \
-                           'See for details' % (package, version)
+                    note = 'Patching or scratch build for %s and version %s FAILED.\n' % (package, version)
```
pnemade: Sorry but to me this English sentence is really difficult to understand. I will only like to suggest drop "and version" words from above. Also, we in general always write or say 
"<name>-<version>" together and not separately.

ralphbean: True there.  So something instead like:

``Patching or scratch build for %s-%s failed." % (...``

I'll go ahead and commit that directly to the ``develop`` branch.

pnemade: :+1:

PR #115, opened 2016-03-16
Handle OSError from 'rm'.

ralphbean:
We're in an exception-handling block here, and, depending on where in the
previous block the exception was raised it is possible that the ``tmp``
directory does not even exist.  So, we have to guard against that and log a
warning.


phracek:
Or may be check before erasing if directory exists.

ralphbean:
> check before erasing if directory exists.

Done!

phracek:
Great. I think that should work properly and better. :+1:

ralphbean:
Thanks :)

PR #124, opened 2016-05-03
Update buildsys.py

phracek:
Argument --buildtool fedpkg is missing.
We need the argument for downloading log, RPMs and other stuff.


pypingou:
Code wise :+1: for me, but I do not know enough the tool to measure its impact, so I trust you on this :)

phracek:
I have tested it today and found this mistake. 
Thanks.

Issue #127, opened 2016-08-12
use mdapi instead of repoquery


ignatenkobrain:
https://pagure.io/mdapi

ignatenkobrain:
I will take care of this.

PR #128, opened 2016-09-03
use mdapi instead of repoquery

ignatenkobrain:
Closes: #127
Signed-off-by: Igor Gnatenko <i.gnatenko.brain@gmail.com>


hotness/consumers.py
```diff
@@ -221,21 +219,19 @@ def _handle_anitya_update(self, upstream, package, msg):
...
-            # likely hasn't yet been a rawhide build of it.. and there's
-            # nothing reasonable we can do.  Notify the world of our failure
-            # and go back to the event loop.
-            self.log.warn("No rawhide version found for %r" % package)
+        mdapi_url = '{0}/srcpkg/koji/{1}'.format(self.mdapi_url, package)
+        self.log.debug("Getting pkg info from %r" % mdapi_url)
+        r = requests.get(mdapi_url)
+        if r.status != 200:
+            # Unfortunately it's not in mdapi, we can't do much about it
+            self.log.warning("No koji version found for %r" % package)
```
pypingou: Should we fall back to the rawhide repo then?

ignatenkobrain: I don't think so. If package not in koji repo, then I don't see point why we should touch rawhide. I guess this is the case for retired packages.

pypingou: Does the koji repo contains all the packages?

ignatenkobrain: I think it does.

bowlofeggs:
This looks good to me, but admittedly I don't know much about mdapi.

pypingou:
I'm going to say, let's merge this but I will not be able to make a release and deploy it before October.

At least this will get this PR out of the queue :)

PR #129, opened 2016-09-26
rebase-helper fixes

nforro:
Be compatible with rebase-helper >= 0.8.
Do not attach logs from rebase-helper as patches.

### DIFF
--- a/hotness/bz.py
+++ b/hotness/bz.py
@@ -122,12 +122,17 @@ class Bugzilla(object):
         self.bugzilla._proxy.Bug.update(update)
         self.log.info("Followed up on bug: %s" % bug.weburl)
 
-    def attach_patch(self, filename, description, bug):
+    def _attach_file(self, filename, description, bug, **kwargs):
         if os.path.exists(filename) and os.path.getsize(filename) != 0:
-            self.log.debug("Attaching patch to bug %r" % bug.bug_id)
-            self.bugzilla.attachfile(bug.bug_id, filename, description,
-                    is_patch=True)
-            self.log.info("Attached patch to bug: %s" % bug.weburl)
+            self.log.debug("Attaching file to bug %r" % bug.bug_id)
+            self.bugzilla.attachfile(bug.bug_id, filename, description, **kwargs)
+            self.log.info("Attached file to bug: %s" % bug.weburl)
+
+    def attach_log(self, filename, description, bug):
+        self._attach_file(filename, description, bug, content_type='text/plain')
+
+    def attach_patch(self, filename, description, bug):
+        self._attach_file(filename, description, bug, is_patch=True)
 
     def ftbfs_bugs(self, name):
         """ Return all FTBFS bugs we find for a package """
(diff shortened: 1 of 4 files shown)


phracek:
@nforro Looks good to me. What about to attach also build.log if SRPM fails?

nforro:
@phracek: This new commit should take care of that.

pypingou:
So am I right to assume rebase-helper 0.8+ has been released?

Does this break backward compatibility? (ie: can I cut a release of the-new-hotness while having a 0.7 version of rebase-helper?)

nforro:
Yes, rebase-helper 0.8 has been released a while ago, it's available in F24 and F25.
And yes, these changes are backward incompatible, so they won't work with rebase-helper 0.7.

hotness/consumers.py
```diff
@@ -283,18 +283,22 @@ def _handle_anitya_update(self, upstream, package, msg):
                 else:
                     note = 'Patching or scratch build for %s-%s failed.\n' % (package, version)
                     self.bugzilla.follow_up(note, bz)
-                    if 'logs' in rh_stuff['build_logs']:
-                        for log in rh_stuff['build_logs']['logs']:
-                            note = 'Build log %s.' % log
-                            self.bugzilla.attach_patch(log, note, bz)
+                    if 'build_logs' in rh_stuff and 'build_ref' in rh_stuff.get('build_logs'):
```
jeremycline: Just a little quibble, but this could be simplified a bit, I think:
```
if 'build_ref' in rh_stuff.get('build_logs', {}):
```
This handles the possibility that 'build_logs' isn't a key in ``rh_stuff`` and provides a default iterable.

hotness/consumers.py
```diff
@@ -283,18 +283,22 @@ def _handle_anitya_update(self, upstream, package, msg):
...
+                        build_ref = rh_stuff['build_logs']['build_ref']
+                        if build_ref:
```
jeremycline: This whole block can be simplified a bit as well. Something like:
```
build_ref = rh_stuff['build_logs']['build_ref'] or {}

for result in build_ref.values():
    for log in result['logs']:
        self.bugzilla.attach_log(log, 'Build log %s.' % log, bz)
```

hotness/bz.py
```diff
@@ -122,6 +122,13 @@ def follow_up(self, text, bug):
         self.bugzilla._proxy.Bug.update(update)
         self.log.info("Followed up on bug: %s" % bug.weburl)
 
+    def attach_log(self, filename, description, bug):
```
jeremycline: It's unfortunate the old function is called ``attach_patch`` since this is basically identical. It might be better to just have new ``attach_file`` function that takes a ``patch`` keyword arg. ``attach_patch`` can then call that with ``patch=True`` and we can mark ``attach_patch`` as deprecated. Does that make sense?

jeremycline:
Oh, one more thing I forgot: since this requires rebase-helper 0.8+, can you update the specfile to require that?

nforro:
Hi Jeremy,
thanks for the comments. In the meantime I refactored and cleaned up most of the rebase-helper related code, but hopefully all your comments should be addresed.
Except the tests, I'm not sure how exactly should I approach them. Any hints would be appreciated.

jeremycline:
@nforro you're definitely correct, the code is in desperate need of a refactor, but I would like to do that separately (and in smaller chunks). Can we keep this PR strictly for making the-new-hotness compatible with rebase-helper-0.8.0?

nforro:
@jeremycline: ok, that makes sense.

jeremycline:
Great, I think we can get tests as we refactor all this stuff.

PR #130, opened 2016-09-27
Fix some bugs in mdapi implementation

nforro:
While testing current the-new-hotness code with local fedmsg instance I found a few issues in mdapi implementation.

@ignatenkobrain: Can you take a look at this?


pypingou:
Let's merge, thanks for the review @ignatenkobrain and the fix @nforro :)

PR #134, opened 2016-11-22
Change the tests to use unittest and add style check

jeremycline:
The tests were using nose and not the standard unittest module. This
changes that and the tests are now runnable by using `python setup.py
test`. This also fixes all PEP8 violations found by the new style test.


pypingou:
:+1: for me

PR #135, opened 2016-11-23
Add timeouts to HTTP requests

jeremycline:
By default, requests does not have a timeout for HTTP requests. This can
cause the application to hang indefinitely if the server never
responses.

### DIFF
--- a/hotness/consumers.py
+++ b/hotness/consumers.py
@@ -22,6 +22,7 @@ Authors:    Ralph Bean <rbean@redhat.com>
 import socket
 import traceback
 
+from requests.packages.urllib3.util import retry
 import fedmsg
 import fedmsg.consumers
 import requests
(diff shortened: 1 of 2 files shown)


hotness/consumers.py
```diff
@@ -220,7 +220,7 @@ def _handle_anitya_update(self, upstream, package, msg):
         # Is it new to us?
         mdapi_url = '{0}/koji/srcpkg/{1}'.format(self.mdapi_url, package)
         self.log.debug("Getting pkg info from %r" % mdapi_url)
-        r = requests.get(mdapi_url)
+        r = requests.get(mdapi_url, timeout=30)
```
pypingou: Would it make sense to have this a configuration variable? Or at least a global variable so we have it stored in one place only?

jeremycline:
You can now configure the connect and read timeouts, as well as the number of times to retry a failed connection. I brought the timeout down to 15 seconds (which is still quite generous in my opinion), but requests are now automatically retried up to 3 times on failure (with exponential backoff between requests) so it'll still make an effort to request something for ~45 seconds

pypingou:
:+1: for me :)

PR #138, opened 2016-11-29
rebase-helper related code refactor

nforro:
Refactor, rewrite and clean up rebase-helper related code.


hotness/buildsys.py
```diff
@@ -163,60 +166,46 @@ def handle(self, package, upstream, version, rhbz):
             shutil.rmtree(tmp)
             pass
 
-    def rebase_helper(self, package, upstream, tmp, bz):
-        """
-        Rebase helper part which does a rebase a inform package
-        maintainer whether package was rebased properly.
-        Output information are in dictionary rh_stuff.
+    def _run_rh(self, package, args, tmp_dir):
```
jeremycline: Could you add a docblock here?

hotness/buildsys.py
```diff
@@ -163,60 +166,46 @@ def handle(self, package, upstream, version, rhbz):
...
+            self.log.info("rebase-helper finished with exit code %d" % result_rh)
             rh_stuff = rh_app.get_rebasehelper_data()
-        except Exception:
-            self.log.exception('Compare packages failed.')
-        self.log.info(rh_stuff)
-
-        return rh_stuff
+            self.log.debug(rh_stuff)
+            return result_rh, rh_stuff
+        finally:
```
jeremycline: This try/finally block is much larger than it needs to be. Can we move this finally up to right after``rh_app.run()``

hotness/consumers.py
```diff
@@ -276,88 +276,74 @@ def _handle_anitya_update(self, upstream, package, msg):
                               "Skipping scratch build.")
                 return
 
-            cwd = os.getcwd()
-            result_rh = 0
-            rh_stuff = {}
-            tmp = tempfile.mkdtemp(prefix='thn-rh', dir='/var/tmp')
```
jeremycline: This function was (and still is) _way_ too large and far too complex. I can see at least 3 nested ``try`` statements. I'd really like to see this broken into several functions.

hotness/consumers.py
```diff
@@ -688,140 +674,104 @@ def get_dist_tag(self):
...
+            return
+
+        if old_task[1] is None or new_task[1] is None:
+            # at least one task is still open
+            return
+
+        bz = old_task[0]
+        upstream = new_task[2]
+
+        if old_task[1] == 'CLOSED':
```
jeremycline: This whole block could probably be its own function

hotness/consumers.py
```diff
@@ -688,140 +674,104 @@ def get_dist_tag(self):
...
+
+        elif old_task[1] == 'FAILED':
+            self.bugzilla.follow_up("Scratch build of rawhide version failed.", bz)
+        elif old_task[1] == 'CANCELED':
+            self.bugzilla.follow_up("Scratch build of rawhide version has been canceled.", bz)
+
+        self.old_triggered_task_ids.pop(old_task_id)
+        self.new_triggered_task_ids.pop(new_task_id)
+
+    def _rh_attach_logs(self, bz, rh_stuff, debug=False):
```
jeremycline: docblock here

hotness/consumers.py
```diff
@@ -688,140 +674,104 @@ def get_dist_tag(self):
...
+            if not log or os.path.getsize(log) == 0:
+                continue
+            notes = ["Log %s from rebase-helper" % os.path.basename(log)]
+            if debug and log.endswith('debug.log'):
+                notes.append("Check it for details and report potential issues "
+                             "to rebase-helper upstream:")
+                notes.append("https://github.com/rebase-helper/rebase-helper/issues")
+            self.bugzilla.attach_log(log, "\n".join(notes), bz)
+
+    def _rh_fail(self, error, bz, rh_stuff):
```
jeremycline: docblock here as well

nforro:
PR updated.
I tried to break down the complex functions as requested. I also added docblocks to all functions.

hotness/consumers.py
```diff
@@ -796,140 +733,230 @@ def get_dist_tag(self):
...
-                                                                            tmp)
-                            bug = bz
-                            self.old_triggered_task_ids.pop(old_task_id)
-                            self.new_triggered_task_ids.pop(task_id)
-                            text = text % (new_state, task_id)
-                            self.bugzilla.follow_up(text, bug)
+
+        old_task_id, new_task_id = self._find_tasks(task_id)
+
+        if old_task_id is None or new_task_id is None:
```
jeremycline: This could also be ``if None in (old_task_id, new_task_id)``, but that's a nitpick and I don't mind either way.

hotness/consumers.py
```diff
@@ -796,140 +733,230 @@ def get_dist_tag(self):
...
+        Try to rebase the package with rebase-helper.
+
+        Runs rebase-helper and follows up on the bugzilla accordingly.
+
+        Returns True unless exception is raised during rebase.
+        """
+        tmp = tempfile.mkdtemp(prefix='thn-rh-', dir='/var/tmp')
+        try:
+            rh_stuff = None
+            try:
```
jeremycline: This nested try block looks unnecessary and you should be able to remove it without affecting the behavior. You can also move the ``rh_stuff = None`` out of the try block so it's clear it's always going to be a variable in the ``except`` block as well.

hotness/consumers.py
```diff
@@ -796,140 +733,230 @@ def get_dist_tag(self):
...
+        new_task = self.new_triggered_task_ids[new_task_id]
+
+        bz = old_task[0]
+        upstream = new_task[2]
+
+        if old_task[1] == 'CLOSED':
+            tmp = tempfile.mkdtemp(prefix='thn-rh-', dir='/var/tmp')
+            try:
+                rh_stuff = None
+                try:
```
jeremycline: Same comment about the nested try block as above

hotness/consumers.py
```diff
@@ -796,140 +733,230 @@ def get_dist_tag(self):
...
+        Returns ids of both tasks.
+        """
+        old_task_id = None
+        new_task_id = None
+
+        for ver in ['old', 'new']:
+            try:
+                task_id = rh_stuff['build_logs']['build_ref'][ver]['koji_task_id']
+                if ver == 'old':
+                    self.old_triggered_task_ids[task_id] = [
```
jeremycline: This list would probably be better as a ``namedtuple`` for the sake of clarity. That way, when it's accessed elsewhere you can do something like ``old_task.bugzilla`` or ``new_task.package``. I like to declare any namedtuples I'm making at the top of the module with docblock.

hotness/consumers.py
```diff
@@ -796,140 +733,230 @@ def get_dist_tag(self):
...
+        Find pair of tasks sharing the same bug id by task id.
+
+        Returns pair of task ids.
+        """
+        old_task_id = None
+        new_task_id = None
+
+        # Find tasks with common bug id
+        if task_id in self.old_triggered_task_ids:
+            old_task_id, old_task = task_id, self.old_triggered_task_ids[task_id]
```
jeremycline: Not that it'd save much, but you could probably define this as a function (maybe just internal to this one) since it's two identical blocks. It's not a big deal either way to me, so just pick whatever you like.

nforro:
@jeremycline: All notes addressed. I used dicts rather than namedtuples, but I can change that if you want.

PR #140, opened 2016-12-05
Add documentation for the message handlers

jeremycline:
This documents the fedmsg consumer class and the handler methods it
implements. There are still several undocumented public functions and
methods, but this covers all the incoming message types and the general
actions taken for each message.


pypingou:
:+1: for me

PR #141, opened 2016-12-08
Add a version to the hotness package

jeremycline:


pypingou:
I remember something about version in module not being handy to import in the setup.py which is why I started putting it there but extracting it by reading the file. IIRC (it was a little while ago) the issue was that by importing the module in the setup.py meant that all the dependencies of the module had to be present before the setup.py could be called, which meant that it broke things like ``python setup.py develop`` in a venv (or something along those lines).

jeremycline:
Oh that's a good point. I've updated the setup.py to pull the version out of the file with regex.

pypingou:
:+1: for me

PR #142, opened 2016-12-08
Modules now log using their own loggers

jeremycline:
Previously, logs nearly all came from the "fedmsg" logger, which is very
confusing. This configures logging for the hotness package and gives
each module its own logger. The timestamp has also been removed by
default since journald keeps timestamps for us. It's easy to add back,
though, and the method is documented in fedmsg.d/logging.py

The old logs looked like:

  [2016-12-06 16:58:28][    fedmsg    INFO] Handling koji...

The new logs look like:

  [hotness.consumers INFO] Handling koji...


pypingou:
:+1: for me

PR #143, opened 2016-12-08
Update usage of the 'userstring' config

jeremycline:
The configuration provided in fedmsg.d/ indicates that 'hotness.koji'
accepts a 'userstring' key that is a tuple of (<name>, <email>). However
the code expects it to be a string.

This modifies hotness/buildsys.py to handle both cases and document the
expected type for this key.

### DIFF
--- a/hotness/buildsys.py
+++ b/hotness/buildsys.py
@@ -1,6 +1,22 @@
-#!/usr/bin/env python
+# -*- coding: utf-8 -*-
 
-from __future__ import print_function
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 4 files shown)


hotness/buildsys.py
```diff
@@ -29,6 +47,12 @@ def __init__(self, consumer, config):
         self.ca_cert = config['ca_cert']
         self.git_url = config['git_url']
         self.userstring = config['userstring']
+        if isinstance(self.userstring, six.string_types):
+            msg = ('Using a string for "userstring" in "hotness.koji"'
+                   ' is deprecated, please use a tuple')
+            warnings.warn(msg, DeprecationWarning)
+        else:
+            self.userstring = ' '.join(self.userstring)
```
pypingou: Are we here converting back the tuple to string?

If that's the case, wouldn't it be more logical to deprecate the support for tuple rather than the one for string?

jeremycline: Yeah, that would make more sense. I kept changing my mind about which one I wanted. I'm still not sure which one I want, but I guess string is easier.

There's a problem where rebase-helper doesn't work because no author is set for the repo (and then the commit fails) and so my motivation for having it be a tuple was that it'd be easier to set the author and email fields. It's probably just as easy to split on ``<`` though. Although I guess someone might want ``<`` in their name (or some reason).

pypingou:
Cool, that even makes the PR simpler, :+1: for me :)

PR #147, opened 2016-12-21
Update the-new-hotness to use Kerberos authentication

jeremycline:
Koji requires Kerberos authentication now, so this switches
the-new-hotness to use that now.

### DIFF
--- a/devel/ansible/roles/hotness-dev/files/fedmsg.d/hotness.py
+++ b/devel/ansible/roles/hotness-dev/files/fedmsg.d/hotness.py
@@ -1,4 +1,3 @@
-import os
 import socket
 hostname = socket.gethostname().split('.', 1)[0]
 
@@ -48,8 +47,14 @@ config = {
     'hotness.koji': {
         'server': 'https://koji.fedoraproject.org/kojihub',
         'weburl': 'http://koji.fedoraproject.org/koji',
-        'cert': os.path.expanduser('~/.fedora.cert'),
-        'ca_cert': os.path.expanduser('~/.fedora-server-ca.cert'),
+        # Kerberos configuration to authenticate with Koji. In development
+        # environments, use `kinit <fas-name>@FEDORAPROJECT.ORG` to get a
+        # Kerberos ticket and set all krb_* settings to `None`.
+        'krb_principal': None,
+        'krb_keytab': None,
+        'krb_ccache': None,
+        'krb_proxyuser': None,
+        'krb_sessionopts': {'timeout': 3600, 'krb_rdns': False},
         'git_url': 'http://pkgs.fedoraproject.org/cgit/rpms/{package}.git',
         'userstring': ('Upstream Monitor',
                        '<upstream-release-monitoring@fedoraproject.org>'),
--- a/hotness/buildsys.py
+++ b/hotness/buildsys.py
@@ -24,6 +24,7 @@ import shutil
 import string
 import subprocess as sp
 import tempfile
+import threading
 import time
 import warnings
 
@@ -37,6 +38,8 @@ from rebasehelper.cli import CLI
 
 _log = logging.getLogger(__name__)
 
+_koji_session_lock = threading.RLock()
+
 
 class Koji(object):
     def __init__(self, consumer, config):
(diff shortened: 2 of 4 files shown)


hotness/buildsys.py
```diff
@@ -60,7 +62,14 @@ def __init__(self, consumer, config):
 
     def session_maker(self):
         koji_session = koji.ClientSession(self.server, {'timeout': 3600})
```
puiterwijk: You need a krb_sessionopts probably, or hardcode {'timeout': 3600} to be {'timeout': 3600, 'krb_rdns': False}.
krb_rdns set to False is a requirement in the Fedora Koji krb auth.

hotness/buildsys.py
```diff
@@ -60,7 +62,14 @@ def __init__(self, consumer, config):
...
-        koji_session.ssl_login(self.cert, self.ca_cert, self.ca_cert)
+        result = koji_session.krb_login(
+            principal=self.krb_principal,
+            keytab=self.krb_keytab,
+            ccache=self.krb_ccache,
+            proxyuser=self.krb_proxyuser
```
puiterwijk: I do not want any new services that use proxyuser, so I'd suggest to make this non-configurable.
If someone wants to use another user, they can just provide that users' principal and keytab.

hotness/buildsys.py
```diff
@@ -60,7 +62,14 @@ def __init__(self, consumer, config):
 
     def session_maker(self):
```
puiterwijk: If you are doing multiple threads at the same time, you want to use a lock in this function, since having multiple threads be in the middle of krb auth at the same time is problematic.

jeremycline:
I added a lock and a ``krb_sessionopts`` config. I didn't remove ``proxyusers`` since it defaults to None and I don't see the harm in allowing someone to configure that if they so choose. We don't have to use it.

devel/ansible/roles/hotness-dev/files/fedmsg.d/hotness.py
```diff
@@ -48,8 +47,14 @@
     'hotness.koji': {
         'server': 'https://koji.fedoraproject.org/kojihub',
         'weburl': 'http://koji.fedoraproject.org/koji',
```
puiterwijk: This should be https://

devel/ansible/roles/hotness-dev/files/fedmsg.d/hotness.py
```diff
@@ -48,8 +47,14 @@
...
-        'ca_cert': os.path.expanduser('~/.fedora-server-ca.cert'),
+        # Kerberos configuration to authenticate with Koji. In development
+        # environments, use `kinit <fas-name>@FEDORAPROJECT.ORG` to get a
+        # Kerberos ticket and set all krb_* settings to `None`.
+        'krb_principal': None,
+        'krb_keytab': None,
+        'krb_ccache': None,
+        'krb_proxyuser': None,
+        'krb_sessionopts': {'timeout': 3600, 'krb_rdns': False},
         'git_url': 'http://pkgs.fedoraproject.org/cgit/rpms/{package}.git',
```
puiterwijk: This should be https://src.fedoraproject.org/...

devel/ansible/roles/hotness-dev/files/fedmsg.d/hotness.py
```diff
@@ -48,8 +47,14 @@
...
         'weburl': 'http://koji.fedoraproject.org/koji',
-        'cert': os.path.expanduser('~/.fedora.cert'),
-        'ca_cert': os.path.expanduser('~/.fedora-server-ca.cert'),
+        # Kerberos configuration to authenticate with Koji. In development
+        # environments, use `kinit <fas-name>@FEDORAPROJECT.ORG` to get a
+        # Kerberos ticket and set all krb_* settings to `None`.
+        'krb_principal': None,
+        'krb_keytab': None,
+        'krb_ccache': None,
+        'krb_proxyuser': None,
```
puiterwijk: I still think this shouldn't be configurable, or at the very least not in the default config, since people might start asking to use it.

jeremycline: I really just put it in there to expose the Koji auth API as-is so people can do whatever they want. I don't know that there are any other deployments of the-new-hotness out there besides ours so maybe this is a moot point, but I erred on the side of being flexible, even though we don't need that flexibility.

jeremycline:
Thanks for the review, I'll update those configuration URLs in a separate PR

devel/ansible/roles/hotness-dev/files/fedmsg.d/hotness.py
pypingou: I am aware of one private instance of anitya, but not for the-new-hotness

PR #148, opened 2017-01-09
Remove rebase-helper

### LINKED ISSUE #145: Consider dropping the dependency on rebase-helper
We've had several problems integrating with [rebase-helper](https://github.com/rebase-helper/rebase-helper) and currently it does not work at all on Fedora 25 in the development environment. I'm quite new to this project so I don't have the historical context of its addition. I've created a list of features rebase-helper appears to add and some challenges I feel it introduces, which I'm certain can be enriched by someone who _does_ have the historical context and knows what drove its addition to the-new-hotness.

Pros
----
1. Removes patches that have already been applied upstream
2. Detect patches that do not apply cleanly
3. Possibly git-apply does a slightly better job of handling patches
4. Runs rpmdiff, pkgdiff, abipkgdiff, csmock on the old and new packages

Of these, I believe 1. could be easily achieved by unpacking the new source and attempting to apply all the patches (patch will complain that the patch was previously applied), 2. is already possibly via a scratch build of the new version and 3. is a minor benefit (in my opinion). 4. is a bit more work.

Cons
----
1. Has made several API-breaking changes and the latest version is not in EPEL. This makes deploying the-new-hotness in production more difficult since it requires wrangling two packages instead of one.
2. Is currently not functional 😢 

It seems to me that we could get most of what rebase-helper offers while removing a good deal of code and a dependency. The addition of rpmdiff, pkgdiff, abipkgdiff, csmock, etc would require some additional work, but it's really about getting all the RPMs (old and new) sorted out and paired up, I think.

What does everyone else think? Am I missing a big feature here?

jeremycline:
This removes the dependency on rebase-helper in the-new-hotness. It also
fixes several of the commands that the-new-hotness shells out to which
depended on git to have a user configured. Finally, it also gets rid of
the 'userstring' setting in favor of a 'email_user' tuple which makes
configuring git with an author simpler.

Fixes #145

### DIFF
--- a/hotness/consumers.py
+++ b/hotness/consumers.py
@@ -20,17 +20,13 @@ Authors:    Ralph Bean <rbean@redhat.com>
 """
 
 import logging
-import os
-import shutil
 import socket
-import tempfile
 import traceback
 
 from requests.packages.urllib3.util import retry
 import fedmsg
 import fedmsg.consumers
 import requests
-import six
 
 import hotness.anitya
 import hotness.buildsys
@@ -176,8 +172,7 @@ class BugzillaTicketFiler(fedmsg.consumers.FedmsgConsumer):
         # Build a little store where we'll keep track of what koji scratch
         # builds we have kicked off.  We'll look later for messages indicating
         # that they have completed.
-        self.old_triggered_task_ids = {}
-        self.new_triggered_task_ids = {}
+        self.scratch_builds = {}
 
         _log.info("That new hotness ticket filer is all initialized")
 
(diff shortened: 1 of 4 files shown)


fedmsg.d/hotness-example.py
```diff
@@ -50,8 +50,7 @@
         'cert': os.path.expanduser('~/.fedora.cert'),
         'ca_cert': os.path.expanduser('~/.fedora-server-ca.cert'),
         'git_url': 'http://pkgs.fedoraproject.org/cgit/rpms/{package}.git',
-        # Previously a tuple was accepted for 'userstring' which is deprecated
-        'userstring': 'Upstream Monitor <upstream-release-monitoring@fedoraproject.org>',
+        'user_email': ('Upstream Monitor', '<upstream-release-monitoring@fedoraproject.org>'),
```
bowlofeggs: Maybe put this in another commit?

hotness/consumers.py
```diff
@@ -329,31 +324,29 @@ def _handle_anitya_update(self, upstream, package, msg):
...
+                # Map that koji task_id to the bz ticket we want to pursue.
+                self.scratch_builds[task_id] = dict(
+                    bz=bz, state=None, version=str(upstream))
+                # Attach the patch to the ticket
+                self.bugzilla.attach_patch(patch_filename, description, bz)
+            except Exception as e:
+                _log.exception(e)
+                note = "An unexpected error occured creating the scratch build:"
+                note = note + "\n\n" + traceback.format_exc()
+                self.bugzilla.follow_up(note, bz)
```
pypingou: We also have this logging module which sends email upon errors, might be useful if you're interested

devel/ansible/roles/hotness-dev/files/fedmsg.d/hotness.py
```diff
@@ -51,8 +51,7 @@
         'cert': os.path.expanduser('~/.fedora.cert'),
         'ca_cert': os.path.expanduser('~/.fedora-server-ca.cert'),
         'git_url': 'http://pkgs.fedoraproject.org/cgit/rpms/{package}.git',
-        'userstring': ('Upstream Monitor',
-                       '<upstream-release-monitoring@fedoraproject.org>'),
+        'user_email': ('Upstream Monitor', '<upstream-release-monitoring@fedoraproject.org>'),
```
jeremycline: I yanked these changes out into #149

hotness/consumers.py
jeremycline: Sounds good. I've changed this to just point the user to the issue tracker and it logs the exception. A later PR can configure the logger to email anything at the exception level (although I think technically this is possible now since we expose a dictionary that's just handed to logging as a dictConfig).

jeremycline:
@bowlofeggs I'd really appreciate a second look-over this as I've done some serious changes breaking it out into two commits.

bowlofeggs:
@jeremycline Hmm, GitHub is showing this PR with one commit. Is the second commit in another PR?

bowlofeggs:
@jeremycline Ah never mind, I see that you explicitly said they were a separate PR in your replies to me.

hotness/consumers.py
```diff
@@ -396,17 +394,18 @@ def handle_buildsys_scratch(self, msg):
 
         url = fedmsg.meta.msg2link(msg, **self.hub.config)
         subtitle = fedmsg.meta.msg2subtitle(msg, **self.hub.config)
-        text2 = "%s %s" % (subtitle, url)
+        text = "%s %s" % (subtitle, url)
```
bowlofeggs: Good change, but I recommend doing it in a separate commit.

PR #152, opened 2017-02-01
Fix tests I broke

jeremycline:
I did bad things and managed to break some tests without noticing. This
addresses all those issues

### DIFF
--- a/hotness/buildsys.py
+++ b/hotness/buildsys.py
@@ -30,9 +30,6 @@ import time
 
 import koji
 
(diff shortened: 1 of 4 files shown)


pypingou:
Looks fine to me

jeremycline:
Thanks for the speedy review, but I think I most have lost changes in a rebase because I missed two unused imports. Anyway, _now_ the tests pass.

pypingou:
Still looking good :)

PR #153, opened 2017-02-01
Exclude tests from being distributed

jeremycline:
This moves the tests into the hotness package and explicitly ignores
them in the setup.py so they aren't included when installed or
distributed.


cverna:
LGTM :)

PR #154, opened 2017-02-10
Gracefully handle removing a directory that's already gone

jeremycline:
If the git clone fails for whatever reason, it removes the target
directory. The finally block is just supposed to make sure the directory
is gone and shouldn't have a problem if it already is gone.


pypingou:
Cool 👍 didn't know this argument, nice one

Issue #160, opened 2017-02-13
The new sources for a package don't actually exist, but we don't report that to the user


jeremycline:
```
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/hotness/consumers.py", line 338, in _handle_anitya_update
    package, upstream, version, bz)
  File "/usr/lib/python2.7/site-packages/hotness/buildsys.py", line 162, in handle
    output = self.run(['spectool', '-g', specfile], cwd=tmp)
  File "/usr/lib/python2.7/site-packages/hotness/buildsys.py", line 121, in run
    raise Exception(message % (' '.join(cmd), p.returncode, out, err))
Exception: cmd:  spectool -g /var/tmp/thn-8gmvbk/lifeograph.spec
return code:  22
stdout:
Getting https://launchpad.net/lifeograph/trunk/1.4.1/+download/lifeograph-1.4.1.rc1.tar.gz to ./lifeograph-1.4.1.rc1.tar.gz
  0     0    0     0    0     0      0      0 --:--:--  0:01:04 --:--:--     0
curl: (22) The requested URL returned error: 404 Not Found
```
If we bump the version and fail to get the new sources, we should report this to the user. The most likely cause for this is, I suspect, Anitya not handling a prefix correctly (a lot of  projects prefix their tags with ``v``, but things like GitHub strip that out in their tarballs and stuff, see https://github.com/release-monitoring/anitya/issues/374)

Issue #158, opened 2017-02-13
Unclear error handling when unable to grab the new sources


jeremycline:
When the-new-hotness sets the version to the latest upstream, its depending on the version being a part of the URL so it can download the new sources. This isn't always the case, and when it's not a ValueError gets raised:
```
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/hotness/consumers.py", line 338, in _handle_anitya_update
    package, upstream, version, bz)
  File "/usr/lib/python2.7/site-packages/hotness/buildsys.py", line 181, in handle
    newfile=newfile, newsum=newsum))
ValueError: spectool was unable to grab new sources

old source: breathe-a423389dfe2935962b3b27586602cfc3521fb54a.tar.gz
old sha256: 135685812242de367cf111ec36a274b786c58356478b60777b1473fb08716f78

new source: ./breathe-a423389dfe2935962b3b27586602cfc3521fb54a.tar.gz
new sha256: 135685812242de367cf111ec36a274b786c58356478b60777b1473fb08716f78
```
We should catch the exception and comment on the bug with instructions on how to fix up the specfile (and a link to our issue tracker in case their specfile is right and something else went wrong).

PR #161, opened 2017-02-15
Rework the buildsys source fetching

### LINKED ISSUE #157: spectool's output can cause an IndexError in buildsys
The traceback:
```
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/hotness/consumers.py", line 338, in _handle_anitya_update
    package, upstream, version, bz)
  File "/usr/lib/python2.7/site-packages/hotness/buildsys.py", line 163, in handle
    newfile = output.strip().split()[-1]
IndexError: list index out of range
```

It looks like the output of ``spectool -g <spec>`` wasn't in the expected format (namely, it didn't have spaces to split on) and this caused an IndexError.

jeremycline:
This improves the error reporting for problems found while fetching the
specfile sources. It should now be more clear to end users what went
wrong and why, and things they can do to improve their specfiles.

Fixes #157, #158, and #160

### DIFF
--- a/hotness/buildsys.py
+++ b/hotness/buildsys.py
@@ -1,5 +1,8 @@
 # -*- coding: utf-8 -*-
-
+#
+# This file is part of the-new-hotness project.
+# Copyright (C) 2017  Red Hat, Inc.
+#
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
 # as published by the Free Software Foundation; either version 2
@@ -18,6 +21,7 @@
 from __future__ import absolute_import, print_function
 
 from warnings import warn
+import hashlib
 import logging
 import os
 import random
@@ -28,7 +32,10 @@ import tempfile
 import threading
 import time
 
+from six.moves.urllib.parse import urlparse
 import koji
+
+from hotness import exceptions
 
 
 _log = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


jeremycline:
To whomever reviews this:

One thing I toyed with a little is dropping the use of ``spectool -g`` to fetch the source tarballs. Instead, we could use ``spectool -l`` to get the source URLs and then use requests to fetch the tarballs. This will make it much easier to provide useful error messages since we wouldn't need to parse the output of curl (on potentially multiple files).

The main reason I didn't immediately go that route is that we _need_ to support FTP (:sadface:). One way to potentially make that less painful is to look into the [requests-ftp](https://github.com/Lukasa/requests-ftp) adapter. It's also worth noting Anitya has this same problem (and I think it currently uses urllib2 directly to deal with FTP) so we might be able to share some code there.

jeremycline:
@ignatenkobrain I only used 256 since that's what I was replacing. I don't mind changing it to 512.

We could (and almost certainly should now that I look at it closely) use pyrpkg directly, but ``fedpkg sources`` was what was being used before (probably because it was fast to implement and it works reasonably well) and I just didn't spend much time investigating how it worked. I'm open to re-writing that, but I didn't want to bite off too much work in this PR.

hotness/buildsys.py
```diff
@@ -200,3 +186,148 @@ def handle(self, package, upstream, version, rhbz):
...
             shutil.rmtree(tmp, ignore_errors=True)
+
+
+def compare_sources(old_sources, new_sources):
+    """
+    Compare two sets of files via checksum and raise an exception if both sets
+    contain the same file.
+
+    Raises:
+        exceptions.SpecUrlException: If the old and new sources share a file
```
bowlofeggs: Maybe docblock the two params.

hotness/buildsys.py
```diff
@@ -200,3 +186,148 @@ def handle(self, package, upstream, version, rhbz):
...
+        raise exceptions.DownloadException(e.output)
+
+    return files
+
+
+def _validate_spec_urls(specfile_path):
+    """
+    Validate a specfile's Source URLs.
+
+    Raises:
```
bowlofeggs: Suggest documenting the param here too.

hotness/buildsys.py
```diff
@@ -200,3 +186,148 @@ def handle(self, package, upstream, version, rhbz):
...
+                   e.output.splitlines()[-1])
+        elif e.returncode == 60:
+            msg = ("Unable to validate the TLS certificate for one of the package's"
+                   "Source URLs")
+        else:
+            msg = (u'An unexpected error occurred while downloading the new package sources; '
+                   u'please report this as a bug on the-new-hotness issue tracker.')
+            _log.error('{cmd} failed (exit {code}): {msg}'.format(
+                cmd=e.cmd, code=e.returncode, msg=e.output))
+        raise exceptions.DownloadException(msg)
```
bowlofeggs: Whoah, very thorough, nice!

hotness/exceptions.py
```diff
@@ -0,0 +1,33 @@
...
+This module contains exceptions raised by the-new-hotness.
+"""
+
+
+class HotnessException(Exception):
+    """The base exception class for all hotness-related errors"""
+
+
+class SpecUrlException(HotnessException):
+    """Raised when a specfile's Source URLs don't work with the-new-hotness"""
```
bowlofeggs: You could make this subclass ValueError too.

hotness/exceptions.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+class DownloadException(HotnessException):
+    """Raised when a networking-related download error occurs"""
```
bowlofeggs: And this could subclass IOError.

Issue #100, opened 2016-02-12
Backand mapping failed for rubygem-concurrent-ruby


voxik:
Not sure how the-new-hotness tries to map the backends, but this is really surprising:

```
hotness tried to map rubygem-concurrent-ruby to an upstream project, but failed:  "Could not determine backend for http://www.concurrent-ruby.com"
    http://www.concurrent-ruby.com
```

Why it tries some URL, when there is `rubygem-` prefix in package name?

jeremycline:
Thanks for the report, @voxik!

This sounds like #116. I know this was filed earlier, but there's more information on #116 so I'm going to close this one in favor of that one.

PR #172, opened 2017-04-18
Detect backend using package name prefix.

### LINKED ISSUE #116: Add guidance on corrective measures when mapping fails
I recently enabled automatic bug+scratch build creation for python-pip in pkgdb, but it failed, sending me the following message:

> hotness tried to map python-pip to an upstream project, but failed:  "Could not determine backend for http://www.pip-installer.org"
>         http://www.pip-installer.org

Presumably that means there's a metadata setting somewhere that needs changing, but I don't know if that's in Anitya, in the spec file, or somewhere else. (Someone that wasn't already familiar with the architecture of the auto-update pipeline would be even more confused)

As a packager, it would likely be very helpful if there was a troubleshooting and/or setup guide walking through the metadata needed to establish the automated update chain, and if that was linked from this error message.

For this particular case, I've checked in Anitya, but python-pip there didn't have that address listed - it had the PyPA docs at https://www.pypa.io/en/latest/

While it didn't match the error message, it still seemed wrong, so I switched it to PyPI's download page, but don't know if there's a way I can force trigger a new fedmsg notification for the latest version.

voxik:
This should resolve #116 (and #100). I implemented only rubygems- prefix support ATM, not sure how exactly other ecosystems works ...

Please note I did not tested the PR at all. I tried to go with Vagrant, but I gave up as soon as it required additional dependencies on my computer (I stumbled up on this already for second time already. Why should I have installed anything except Vagrant on my computer? I thought that Vagrant should shield me of all this).

### DIFF
--- a/hotness/anitya.py
+++ b/hotness/anitya.py
@@ -81,6 +81,43 @@ def _parse_service_form(response):
     return (parsed.form.attrs['action'], inputs)
 
 
+def determine_backend(project_name, project_homepage):
+    """
+    Determine the Anitya backend to use for a given project name and homepage.
+
+    This is not a 100% accurate process. If the ``project_name`` has a prefix with
+    a backend associated with it in :data:`prefixes`, that is preferred. Otherwise,
+    the homepage is checked to see if the host is mapped to known backend.
+
+    Args:
+        project_name (str): The project's name, possibly containing the RPM prefix
+            (e.g. ``python-`` for Python packages).
+        project_homepage (str): The project's homepage URL.
+
+    Returns:
+        str: The backend name to use with Anitya.
+
+    Raises:
+        AnityaException: When the backend could not be determined.
+    """
+    for prefix, backend in prefixes.items():
+        if project_name.startswith(prefix) and backend is not None:
+            return backend
+
+    for target, backend in backends.items():
+        if target in project_homepage:
+            return backend
+
+    err = ("the-new-hotness was unable to automatically determine the Anitya backend "
+           "to use with the {name} ({homepage}) project. Please search {anitya_url} "
+           "for this project. If it already exists, ensure there is a distribution "
+           "mapping for Fedora. If it does not already exist, please create the "
+           "project manually. Doing so enables us to automatically notify package "
+           "maintainers when new versions are released.").format(
+                name=project_name, homepage=project_homepage, anitya_url=ANITYA_URL)
+    raise AnityaException(err)
+
+
 class Anitya(OpenIdBaseClient):
     def __init__(self, url=ANITYA_URL, insecure=False):
         super(Anitya, self).__init__(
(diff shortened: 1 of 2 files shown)


hotness/anitya.py
```diff
@@ -215,10 +215,18 @@ def add_new_project(self, name, homepage):
...
+            if name.startswith(prefix) and backend != None:
                 data['backend'] = backend
-                break
+
+        # And try homepage if backend was not detected based on prefix
+        if 'backend' not in data:
+            for target, backend in backends.items():
+                if target in homepage:
+                    data['backend'] = backend
+                    break
```
bowlofeggs: I might recommend trying the homepage before trying the backend though I could see an argument for doing it this way. As a packager, if I set the home page to the GitHub page, I might want GitHub tags to be the thing that is used. Also, not all ```python-``` packages that are on GitHub would necessarily be on PyPI (for example, I know this PR is just for Ruby for now).

voxik:
You expressed precisely why I thought I'll try the stuff in Vagrant, just to find out that I would need to install Ansible directly on my computer ....

hotness/anitya.py
voxik: Almost 100% gems are available on RubyGems, while probably 70% are hosted on GitHub and may be 10% of them, although hosted on GitHub has different homepage. So for RubyGems, trying homepage first is just waste of energy.

Also, the git tags are not standardized and people are using different formats, that is another downside of GitHub backend ...

jeremycline: Hmmm. I wonder if it'd be best to try the prefix-determined backend first and prod the location with a request, falling back to the homepage if that's an invalid URL. Although this wouldn't be perfect (plenty of projects neglect their PyPI releases, for example), it'd be even better.

I've been slowly working towards breaking the backend of Anitya out into a separate package that we can then use in the-new-hotness, so we could leverage the actual backend code in Anitya. Another thought I just had is it might be nice to have a little flag on projects in Anitya to indicate they were generated by a machine and might not be right.

Anyway, all that's to say that I think this is a reasonable default for the moment.

jeremycline:
Thanks for the PR, @voxik!

I think the change looks reasonable, but it needs tests. I'm happy to write them myself, but it may be a little while until I get to it.

As for Vagrant + Ansible, I think it's a reasonable trade-off for a reliable development environment that I'm happy to support if you run into trouble. However, you're free to set this up in a virtualenv if you'd prefer. You'll have to install things though, since (for the moment) it requires the RPM bindings which aren't on PyPI.

voxik:
It would be nice if Ansible was installed on the guest and executed on the guest .... Not sure if that is possible, but that is how I could avoid installation of anything except Vagrant.

hotness/anitya.py
```diff
@@ -81,6 +81,43 @@ def _parse_service_form(response):
...
+    for target, backend in backends.items():
+        if target in project_homepage:
+            return backend
+
+    err = ("the-new-hotness was unable to automatically determine the Anitya backend "
+           "to use with the {name} ({homepage}) project. Please search {anitya_url} "
+           "for this project. If it already exists, ensure there is a distribution "
+           "mapping for Fedora. If it does not already exist, please create the "
+           "project manually. Doing so enables us to automatically notify package "
+           "maintainers when new versions are released.").format(
```
jeremycline: @ncoghlan what do you think about this in terms of guidance for #116?

ncoghlan:
Regarding Vagrant+Ansible: running the provisioning system from the host is just part of how Vagrant works.

You can use a virtualenv to run Ansible if you don't otherwise have it installed, but if you're going to use virtualenv for Anitya development anyway, it's easier to just skip using Vagrant entirely.

hotness/anitya.py
```diff
@@ -33,17 +33,17 @@
     'sourceforge.net': 'Sourceforge',
```
ncoghlan: Tangent: `pypi.org` should be added as a second target domain for PyPI

hotness/anitya.py
ncoghlan: Looks good to me.

PR #178, opened 2017-06-19
Report SRPM build errors to users

### LINKED ISSUE #171: Unexpected error
hotness told me to do this:

https://bugzilla.redhat.com/show_bug.cgi?id=1435488

jeremycline:
At the moment, hotness shells out a lot and any time something goes
wrong a base ``Exception`` is raised. This needs to be re-written, but
in the meantime, this will at least let users know when and why an SRPM
failed to build.

Additionally, the generic exception message has been changed because it does actually email me for all the errors.

There aren't any tests (for which I am ashamed), but I want to re-write all this code and the only thing the tests could really do at the moment are assert I'm shelling out and ``Exception`` is raised. 😞 

fixes #171

### DIFF
--- a/hotness/consumers.py
+++ b/hotness/consumers.py
@@ -346,9 +346,8 @@ class BugzillaTicketFiler(fedmsg.consumers.FedmsgConsumer):
                 self.bugzilla.follow_up(str(e), bz)
             except Exception as e:
                 _log.exception(e)
-                note = ("An unexpected error occured creating the scratch build: "
-                        "please report this issue to the-new-hotness issue tracker "
-                        "at https://github.com/fedora-infra/the-new-hotness/issues")
+                note = ("An unexpected error occurred while creating the scratch build "
+                        "and has been automatically reported. Sorry!")
                 self.bugzilla.follow_up(note, bz)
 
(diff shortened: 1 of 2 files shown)


hotness/consumers.py
```diff
@@ -346,9 +346,8 @@ def _handle_anitya_update(self, upstream, package, msg):
                 self.bugzilla.follow_up(str(e), bz)
             except Exception as e:
                 _log.exception(e)
-                note = ("An unexpected error occured creating the scratch build: "
-                        "please report this issue to the-new-hotness issue tracker "
-                        "at https://github.com/fedora-infra/the-new-hotness/issues")
+                note = ("An unexpected error occurred creating the scratch build "
```
pypingou: occurred creating -> occurred while creating ?

pypingou:
One comment, otherwise looks fine

hotness/consumers.py
jeremycline: Hmm. I _think_ both are valid English (or at least, they pass my internal "that sounds right" checks). I'm guessing the second version sounds better to non-native speakers so I'll change it.

PR #185, opened 2017-08-30
Port away from pkgdb.

ralphbean:
Still in progress.  Started adding tests.

### DIFF
--- a/hotness/consumers.py
+++ b/hotness/consumers.py
@@ -392,7 +392,7 @@ class BugzillaTicketFiler(fedmsg.consumers.FedmsgConsumer):
             return False
 
         url = '{0}/raw/master/f/rpms/{1}'.format(self.repo_url, package)
-        _log.debug("Checking %r" % url)
+        _log.debug("Checking %r to see if %s is monitored." % (url, package))
         r = self.requests_session.get(url, timeout=self.timeout)
 
         if not r.status_code == 200:
@@ -420,7 +420,8 @@ class BugzillaTicketFiler(fedmsg.consumers.FedmsgConsumer):
             type='rpm',
             active=True,
         )
-        _log.debug("Checking %r, %r" % (url, params))
+        _log.debug("Checking %r to see if %s is retired, %r" % (
+            url, package, params))
         r = self.requests_session.get(url, params=params, timeout=self.timeout)
 
         if not r.status_code == 200:
@@ -436,7 +437,7 @@ class BugzillaTicketFiler(fedmsg.consumers.FedmsgConsumer):
         """ Returns True if a package is in the Fedora dist-git. """
 
         url = '{0}/rpms/{1}'.format(self.dist_git_url, package)
-        _log.debug("Checking %r" % url)
+        _log.debug("Checking %r to see if %s is in dist-git" % (url, package))
         r = self.requests_session.head(url, timeout=self.timeout)
 
         if r.status_code == 404:
--- a/hotness/consumers.py
+++ b/hotness/consumers.py
@@ -26,6 +26,7 @@ import subprocess
 from requests.packages.urllib3.util import retry
 import fedmsg
 import fedmsg.consumers
+import fedmsg.meta
 import requests
 
 from hotness import exceptions
@@ -123,7 +124,8 @@ class BugzillaTicketFiler(fedmsg.consumers.FedmsgConsumer):
 
         # First, initialize fedmsg and bugzilla in this thread's context.
         hostname = socket.gethostname().split('.', 1)[0]
-        fedmsg.init(name='hotness.%s' % hostname)
+        if not getattr(getattr(fedmsg, '__local', None), '__context', None):
+            fedmsg.init(name='hotness.%s' % hostname)
         fedmsg.meta.make_processors(**self.hub.config)
 
         self.bugzilla = hotness.bz.Bugzilla(
--- a/hotness/consumers.py
+++ b/hotness/consumers.py
@@ -28,6 +28,7 @@ import fedmsg
 import fedmsg.consumers
 import fedmsg.meta
 import requests
+import yaml
 
 from hotness import exceptions
 import hotness.anitya
@@ -135,6 +136,10 @@ class BugzillaTicketFiler(fedmsg.consumers.FedmsgConsumer):
 
         default = 'https://admin.fedoraproject.org/pkgdb/api'
         self.pkgdb_url = self.config.get('hotness.pkgdb_url', default)
+        default = 'https://pagure.io/releng/fedora-scm-requests'
+        self.repo_url = self.config.get('hotness.repo_url', default)
+        default = 'https://pdc.fedoraproject.org'
+        self.pdc_url = self.config.get('hotness.pdc_url', default)
 
         anitya_config = self.config.get('hotness.anitya', {})
         default = 'https://release-monitoring.org'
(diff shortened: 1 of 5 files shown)


ralphbean:
OK, I think this is ready for review.

Note that the flags it looks for in the git repo should all be there after this mega commit that migrated the data from old pkgdb:  https://pagure.io/releng/fedora-scm-requests/c/d468d8513eaebfc41a4f37c1549f7a3ebf319750?branch=master

The tests pass locally and I functionally verified that a spoofed [anitya.project.version.update](https://apps.fedoraproject.org/datagrepper/id?id=2017-76c8deb8-2fac-46f1-8287-422ed02b38e4&is_raw=true&size=extra-large) message was [handled correctly](https://partner-bugzilla.redhat.com/show_bug.cgi?id=1475090).

hotness/consumers.py
```diff
@@ -123,16 +105,21 @@ def __init__(self, hub):
...
         fedmsg.meta.make_processors(**self.hub.config)
 
         self.bugzilla = hotness.bz.Bugzilla(
             consumer=self, config=self.config['hotness.bugzilla'])
         self.buildsys = hotness.buildsys.Koji(
             consumer=self, config=self.config['hotness.koji'])
 
-        default = 'https://admin.fedoraproject.org/pkgdb/api'
-        self.pkgdb_url = self.config.get('hotness.pkgdb_url', default)
+        default = 'https://pagure.io/releng/fedora-scm-requests'
```
jeremycline: This is a pattern I've observed throughout Fedora infra code which I really don't like. 

There are a couple problems with this approach. The first problem is that often a configuration value is used in more than just one place, and can result in two (or more!) different defaults in different places in the code. The second problem is that if one wants to know what all the default values for various configurations are (or just know what configuration options are used), you have to comb the code to look for this sort of stuff.

Since you're just refactoring code that already did this, I'm okay with this being accepted as-is, but we need to file an issue about adding a configuration module that has all the valid keys and default values for them. It is also a little weird that you repeatedly overwrite ``default``. Why not just pass the string literal to ``get``?

hotness/consumers.py
```diff
@@ -698,46 +412,50 @@ def is_monitored(self, package):
...
+    @logged
+    def is_retired(self, package):
+        url = '{0}/rest_api/v1/component-branches/'.format(self.pdc_url)
+        params = dict(
+            name='master',
+            global_component=package,
+            type='rpm',
+            active=True,
+        )
+        _log.debug("Checking %r, %r" % (url, params))
```
jeremycline: It'd be nice if this was more informative. Something like
```
'Querying {url} to determine if {package} is retired'
```
And then some sort of followup with the result.

Alternatively, you could make the request first and log in a single statement the result.

hotness/consumers.py
```diff
@@ -698,46 +412,50 @@ def is_monitored(self, package):
...
+        r = self.requests_session.get(url, params=params, timeout=self.timeout)
```
jeremycline: This could raise a number of different exceptions. You need to either declare those in the documentation for this function or handle them here.

hotness/consumers.py
```diff
@@ -698,46 +412,50 @@ def is_monitored(self, package):
...
+                'no-monitoring': False,
+            }
+            return return_values[string_value]
         except:
-            _log.exception("Problem interacting with pkgdb.")
+            _log.exception("Problem interacting with the pagure repo.")
             return False
 
+    @logged
+    def is_retired(self, package):
```
jeremycline: Please document this method and its API.

hotness/tests/test_consumer.py
```diff
@@ -0,0 +1,130 @@
+"""
+Unit tests for hotness.consumer
+"""
+from __future__ import unicode_literals, absolute_import
+
+import unittest
+
+import mock
+
+import fedmsg.config
```
jeremycline: This should be grouped with 3rd-party libraries (mock)

hotness/tests/test_consumer.py
```diff
@@ -0,0 +1,130 @@
...
+        test_config["hotness.cache"] = {
+            "backend": "dogpile.cache.null",
+        }
+        test_config['hotness.requests_retries'] = 0
+
+        class MockHub(mock.MagicMock):
+            config = test_config
+
+        self.consumer = consumers.BugzillaTicketFiler(MockHub())
+        import logging
```
jeremycline: Please have all imports at the top of the module

hotness/tests/test_consumer.py
```diff
@@ -0,0 +1,130 @@
...
+        logging.basicConfig(level=logging.DEBUG)
+
+    def tearDown(self):
+        self.bz.__exit__()
+        self.koji.__exit__()
+
+    def test_is_retired_negative(self):
```
jeremycline: Any reason you opted to not use vcrpy to record real responses? I know you mentioned it at FLOCK.

hotness/tests/test_consumer.py
```diff
@@ -0,0 +1,130 @@
...
+        response.json.return_value = {'count': 0}
+        self.consumer.requests_session = mock.MagicMock()
+        self.consumer.requests_session.get.return_value = response
+
+        package = 'python-sqlite2'
+        expected = True
+        actual = self.consumer.is_retired(package)
+        self.assertEquals(expected, actual)
+
+    def test_is_retired_positive(self):
```
jeremycline: I find the test names a little confusing when combined with ``is_retired`` (which could probably just be called ``retired``). To me positive implies "Yes, it _is_ retired". I don't mind one way or the other since the docblock clears it up, but just so you know.

hotness/tests/test_consumer.py
```diff
@@ -0,0 +1,130 @@
...
+        response.json.return_value = {'count': 1}
+        self.consumer.requests_session = mock.MagicMock()
+        self.consumer.requests_session.get.return_value = response
+
+        package = 'nethack'
+        expected = False
+        actual = self.consumer.is_retired(package)
+        self.assertEquals(expected, actual)
+
+    def test_is_monitored_negative(self):
```
jeremycline: Adding a short docblock explaining what you intend to test is very helpful for when the tests do their job and break.

hotness/consumers.py
```diff
@@ -39,6 +41,17 @@
 _log = logging.getLogger(__name__)
 
 
+def logged(fn):
```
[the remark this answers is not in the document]
ralphbean: Removed in c28ce46

hotness/consumers.py
ralphbean: Agreed.  Filed as #186.

hotness/consumers.py
ralphbean: Added some more context in 92eeeb3.

hotness/consumers.py
ralphbean: Agreed.  I added a generic warning in 0d818b8.

hotness/tests/test_consumer.py
ralphbean: Agreed.  Fixed in 79d4ec3.

hotness/tests/test_consumer.py
ralphbean: Moved in b3de423.

hotness/tests/test_consumer.py
ralphbean: Yeah - I just had trouble getting it to work and then bailed for lack of time.

hotness/tests/test_consumer.py
ralphbean: Yeah - this is confusing.  Will address.

ralphbean: Adjusted in a5bfa03.

hotness/tests/test_consumer.py
ralphbean: Added in 7fc255e.

hotness/consumers.py
```diff
@@ -26,7 +26,9 @@
 from requests.packages.urllib3.util import retry
 import fedmsg
 import fedmsg.consumers
+import fedmsg.meta
 import requests
+import yaml
```
jeremycline: Sorry I missed this first go around, could you please add this dependency to the requirements.txt?

ralphbean: Oh, for sure.

ralphbean: Done in 9c97fcf.

ralphbean:
Ready for review again, @jeremycline.

jeremycline:
Perfect, thanks!

PR #197, opened 2018-09-26
Update docs

Zlopez:

PR #199, opened 2018-09-26
Update vagrant container

Zlopez:
This PR takes in consideration https://github.com/fedora-infra/the-new-hotness/pull/198 and https://github.com/fedora-infra/the-new-hotness/pull/197.

These two should be merged first.

### DIFF
--- a/devel/ansible/roles/hotness-dev/files/motd
+++ b/devel/ansible/roles/hotness-dev/files/motd
@@ -1,14 +1,27 @@
 
 Welcome to The New Hotness development environment!
 
+To start development you need to do this first:
+
+* add your bugzilla credentials to ~/.fedmsg.d/hotness.py
+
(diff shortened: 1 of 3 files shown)


devel/ansible/roles/hotness-dev/files/motd
```diff
@@ -1,14 +1,27 @@
 
 Welcome to The New Hotness development environment!
 
+To start development you need to do this first:
+
+* add your bugzilla credentials to ~/.fedmsg.d/hotness.py
+
+* acquire a valid Kerberos ticket by `kinit <fas-username>@FEDORAPROJECT.ORG`
```
bowlofeggs: I recommend using the staging environment instead of production if that's feasible. Bodhi uses production and I've been meaning to switch it to staging to prevent any accidental connections to production.

Zlopez: I want to only simulate the connection for development environment. But you are right about this.

PR #212, opened 2019-01-29
Migrate to fedora-messaging

### LINKED ISSUE #211: Convert to fedora-messaging
Because fedmsg will be replaced by fedora-messaging. We should integrate fedora-messaging to the-new-hotness.

Zlopez:
This change covers:
* code - publisher and consumer
* vagrant development environment
* documentation
* configuration
* basic schema that is now validating current fedmsg messages

Fixes #211


config/config.toml.example
```diff
@@ -0,0 +1,167 @@
...
+durable = true
+auto_delete = false
+exclusive = false
+arguments = {}
+
+[queues.koji]
+durable = true
+auto_delete = false
+exclusive = false
+arguments = {}
```
abompard: Any reason to define two separate queues here? I think you could define only one, call it "the-new-hotness" for example (I don't remember if dashes are allowed), and bind all the topics to it.

Queues are usually named after the consumer that read them, not after the producer.

config/config.toml.example
```diff
@@ -0,0 +1,167 @@
...
+
+[tls]
+ca_cert = "/etc/pki/tls/certs/ca-bundle.crt"
+keyfile = "/my/client/key.pem"
+certfile = "/my/client/cert.pem"
+
+[client_properties]
+app = "the-new-hotness"
+
+[exchanges.custom_exchange]
```
abompard: Why define this exchange?

devel/ansible/roles/hotness-dev/tasks/main.yml
```diff
@@ -1,39 +1,47 @@
...
-    - python3-sh
-    - python3-six
-    - python-fedmsg-meta-fedora-infrastructure
-    - python3-fedmsg
-    - rpmdevtools
+  dnf:
+    name: [
+      fedmsg,
+      fedmsg-hub,
+      fedmsg-relay,
```
abompard: Do you still need those 3?

devel/ansible/roles/hotness-dev/tasks/main.yml
```diff
@@ -1,39 +1,47 @@
...
+      rpm-python,
+      python3-bugzilla,
+      python3-dogpile-cache,
+      python3-flake8,
+      python3-mock,
+      python3-requests,
+      python3-sh,
+      python3-six,
+      python-fedmsg-meta-fedora-infrastructure,
+      python3-fedmsg,
```
abompard: Likewise, do you still need those two?

config/config.toml.example
```diff
@@ -0,0 +1,167 @@
...
+# before giving up.
+connect_timeout = 15
+# The time in seconds the-new-hotness should wait for a read from a socket
+# before giving up.
+read_timeout = 15
+# The number of times the-new-hotness should retry a network request that
+# that failed for any reason (e.g. read timeout, DNS error, etc)
+requests_retries = 3
+# If true, publish fedmsg messages instead of fedora-messaging messages
+legacy_messaging = false
```
abompard: Hmm OK but you're still only *consuming* from fedora-messaging, and not from fedmsg, right?

requirements.txt
```diff
@@ -8,3 +8,4 @@ PyYAML
 koji
 psutil
 moksha.hub
```
abompard: You probably don't need this one anymore.

config/config.toml.example
Zlopez: I named them after producer, but this is probably not good approach. I will change this.

config/config.toml.example
Zlopez: I probably copied it from some example in fedora-messaging. I will remove it.

devel/ansible/roles/hotness-dev/tasks/main.yml
Zlopez: It's question if we want to remove the fedmsg in one go or leave the fedmsg for while.

devel/ansible/roles/hotness-dev/tasks/main.yml
Zlopez: Same as above. I would be rather to remove this at once, but I don't want to end up with the-new-hotness not working, because I removed something that was needed.

config/config.toml.example
Zlopez: Yes, this is only for publishing.

requirements.txt
Zlopez: Probably.

devel/ansible/roles/hotness-dev/tasks/main.yml
Zlopez: This is in the dev environment, so it's safe to remove it.

config/config.toml.example
```diff
@@ -0,0 +1,155 @@
...
+[[bindings]]
+queue = "anitya"
+exchange = "amq.topic"
+routing_keys = [
+    "org.release-monitoring.prod.anitya.project.version.update",
+    "org.release-monitoring.prod.anitya.project.map.new",
+]
+
+[[bindings]]
+queue = "koji"
```
abompard: You need to update this to `the-new-hotness`.

config/config.toml.example
```diff
@@ -0,0 +1,155 @@
...
+amqp_url = "amqp://"
+
+publish_exchange = "amq.topic"
+
+callback = "hotness.consumers:BugzillaTicketFiler"
+
+# Note the double brackets below.
+# To add another binding, add another [[bindings]] section.
+[[bindings]]
+queue = "anitya"
```
abompard: You need to update this to `the-new-hotness`.

devel/ansible/roles/hotness-dev/files/config.toml
```diff
@@ -0,0 +1,155 @@
...
+amqp_url = "amqp://"
+
+publish_exchange = "amq.topic"
+
+callback = "hotness.consumers:BugzillaTicketFiler"
+
+# Note the double brackets below.
+# To add another binding, add another [[bindings]] section.
+[[bindings]]
+queue = "anitya"
```
abompard: You need to update this to `the-new-hotness`.

devel/ansible/roles/hotness-dev/files/config.toml
```diff
@@ -0,0 +1,155 @@
...
+exchange = "amq.topic"
+routing_keys = [
+    "org.release-monitoring.prod.anitya.project.version.update",
+    "org.release-monitoring.prod.anitya.project.map.new",
+]
+
+[[bindings]]
+queue = "koji"
```
abompard: You need to update this to `the-new-hotness`.

config/config.toml.example
```diff
@@ -0,0 +1,155 @@
...
+    "org.release-monitoring.prod.anitya.project.version.update",
+    "org.release-monitoring.prod.anitya.project.map.new",
+]
+
+[[bindings]]
+queue = "then-new-hotness"
+exchange = "amq.topic"
+routing_keys = [
+    "org.fedoraproject.prod.buildsys.task.state.change",
+]
```
abompard: Having 2 `bindings` sections like that will work, but you may want to add the second routing key to the first list to get the same result.

devel/ansible/roles/hotness-dev/files/config.toml
```diff
@@ -0,0 +1,155 @@
...
+    "org.release-monitoring.prod.anitya.project.version.update",
+    "org.release-monitoring.prod.anitya.project.map.new",
+]
+
+[[bindings]]
+queue = "the-new-hotness"
+exchange = "amq.topic"
+routing_keys = [
+    "org.fedoraproject.prod.buildsys.task.state.change",
+]
```
abompard: Same remark here.

devel/ansible/roles/hotness-dev/tasks/main.yml
```diff
@@ -45,14 +48,24 @@
...
-    src: fedmsg.d/
-    dest: /home/{{ ansible_env.SUDO_USER }}/.fedmsg.d
+    src: config.toml
+    dest: /home/{{ ansible_env.SUDO_USER }}/
+
+- name: Enable RabbitMQ server
+  command: systemctl enable rabbitmq-server
+
+- name: Start RabbitMQ server
+  command: systemctl start rabbitmq-server
```
abompard: You could also use [the systemd module](https://docs.ansible.com/ansible/latest/modules/systemd_module.html#systemd-module) for those two.

hotness/consumers.py
```diff
@@ -160,17 +127,35 @@ def __init__(self, hub):
...
+            topic (str): Topic to publish to
+            msg (dict): Message to be send
         """
         _log.info("publishing topic %r" % topic)
-        fedmsg.publish(modname="hotness", topic=topic, msg=msg)
+        if self.config["legacy_messaging"]:
+            fedmsg.publish(modname="hotness", topic=topic, msg=msg)
+        else:
+            try:
+                fm_publish(Message(topic=topic, body=msg))
```
abompard: Actually, fedmsg mangles the topic you give it and prefixes the modname value. To get the equivalent in fedora-messaging, you must use `"hotness.{}".format(topic)` for the topic argument.

hotness_schema/hotness_schema/messages.py
```diff
@@ -0,0 +1,171 @@
...
+
+        Returns:
+            (str): Short description of the message.
+        """
+        return "the-new-hotness filed a bug on '{}'".format(self.package)
+
+    @property
+    def package(self):
+        """
+        Return a reason for this drop.
```
abompard: This docstring seems wrong.

hotness_schema/setup.py
```diff
@@ -0,0 +1,60 @@
...
+from setuptools import setup, find_packages
+
+
+here = os.path.abspath(os.path.dirname(__file__))
+with open(os.path.join(here, "README.rst")) as fd:
+    README = fd.read()
+
+
+setup(
+    name="the-new-hotness_schema",
```
abompard: We're trying to standardize the names a bit, I would suggest `fedora_messaging_the_new_hotness_schemas`. I know it's a lot of underscores, but I would recommend against mixing underscores and dashes.

devel/ansible/roles/hotness-dev/tasks/main.yml
Zlopez: Good point

hotness/consumers.py
Zlopez: Ok, didn't knew this.

hotness_schema/setup.py
Zlopez: As you wish, sir :-)

abompard:
OK, this looks good! However I'm a bit worried that the difference in topic format from fedmsg to fedora-messaging wasn't caught by any unit tests. Could there be something missing here or do you think it's not a problem?
If you're OK with this and the changeset as it is, feel free to commit (probably after squashing those commits, but it's up to you).

Zlopez:
the-new-hotness doesn't have much tests. I want to change this in the future, but right now the publish method doesn't have any test at all.

abompard:
Alright then, LGTM :-)

PR #225, opened 2019-03-11
Fix hotness fedora messaging schema

### LINKED ISSUE #224: Fix update.drop schema
the-new-hotness fedora messaging messages are not the same as fedmsg. We must fix this because it's causing issue in `fmn` when trying to work with `fedmsg_meta_fedora_infrastructure`:
```
File "/usr/lib/python2.7/site-packages/fedmsg_meta_fedora_infrastructure/hotness.py", line 223, in usernames
  if 'buildsys.build' in msg['msg']['trigger']['topic']:
```

We need to send the message as it is expected to be by `fedmsg_meta_fedora_infrastructure`.

Zlopez:
This should fix the issue with fedmsg_meta_fedora_infrastructure. Which
had trouble parsing the current messages because of the missing values
in trigger object.

I used the same logic for the summary methods as in [fedmsg_meta_fedora_infrastructure](https://github.com/fedora-infra/fedmsg_meta_fedora_infrastructure/blob/develop/fedmsg_meta_fedora_infrastructure/hotness.py).

Fixes #224 

### DIFF
--- a/hotness_schema/hotness_schema/messages.py
+++ b/hotness_schema/hotness_schema/messages.py
@@ -35,32 +35,11 @@ class UpdateDrop(message.Message):
             "reason": {"type": "string"},
             "trigger": {
                 "type": "object",
-                "properties": {
-                    "msg": {
-                        "type": "object",
-                        "properties": {
-                            "project": {
-                                "type": "object",
-                                "properties": {"name": {"type": "string"}},
-                                "required": ["name"],
-                            }
-                        },
-                        "required": ["project"],
-                    }
-                },
-                "required": ["msg"],
+                "properties": {"msg": {"type": "object"}, "topic": {"type": "string"}},
+                "required": ["msg", "topic"],
             },
         },
     }
-
-    def __str__(self):
-        """
-        Return a complete human-readable representation of the message.
-
-        Returns:
-            (str): Summary of the message.
-        """
-        return self.summary
 
     @property
     def summary(self):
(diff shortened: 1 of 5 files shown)


hotness_schema/hotness_schema/messages.py
```diff
@@ -72,39 +51,68 @@ def summary(self):
...
             return "the-new-hotness saw an update for '{}', ".format(
-                self.project
+                self.package
             ) + "but it got dropped for reason: '{}'".format(self.reason)
 
     @property
-    def project(self):
+    def package(self):
         """
         Return a name of the project.
```
abompard: The docstring may need to be updated.

hotness_schema/hotness_schema/messages.py
```diff
@@ -158,14 +161,24 @@ def summary(self):
...
     @property
-    def package(self):
+    def packages(self):
         """
-        Return package name.
+        Return package names.
 
         Returns:
-            (str): Package name.
+            (str): Package names delimited by ','.
```
abompard: The signature of the `packages()` method on the `Message` class is to return a list.

hotness_schema/hotness_schema/messages.py
```diff
@@ -72,39 +51,68 @@ def summary(self):
...
+            return original["package_listing"]["package"]["name"]
+
+        if "buildsys.build" in self.body["trigger"]["topic"]:
+            return self.body["trigger"]["msg"]["name"]
+
+        if "package" in self.body["trigger"]["msg"]:
+            original = self.body["trigger"]["msg"]
+            return original["package"]["name"]
+
+        return ""
```
abompard: Are you sure you want to return an empty string if the package can't be found? Returning `None` seems more appropriate, but it's up to you.

Also, it'd be nice to redefine the `packages()` method to return a list containing only the result of the `package()` method (if not empty / not `None`)

hotness_schema/hotness_schema/messages.py
Zlopez: I didn't noticed, that `Message` class has already `packages()` method. I will fix this.

hotness_schema/hotness_schema/messages.py
Zlopez: This is the same logic, that the fedmsg_meta_fedora_infrastructure is using.

It will be probably better to override `packages()` property from `Message`.

Zlopez:
Thanks, but it will not be merged anyway. There is still issue with CI.

PR #236, opened 2019-05-24
Fix hijacking bugzilla issue

### LINKED ISSUE #234: the-new-hotness override a wrong issue in RHBZ
A new release report of deepin-terminal from Anitya overrided an ABRT issue:
https://bugzilla.redhat.com/show_bug.cgi?id=1688632
which is private.

EDIT BY puiterwijk: Posting screenshots of private tickets is *not* a good idea.

Zlopez:
Add reporter to bugzilla query

Fixes #234

### DIFF
--- a/devel/ansible/roles/hotness-dev/files/config.toml
+++ b/devel/ansible/roles/hotness-dev/files/config.toml
@@ -55,7 +55,7 @@ propagate = false
 handlers = ["console"]
 
 [log_config.root]
-level = "WARNING"
+level = "DEBUG"
 handlers = ["console"]
 
 # the-new-hotness consumer configuration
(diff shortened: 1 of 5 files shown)


devel/ansible/roles/hotness-dev/files/config.toml
```diff
@@ -55,7 +55,7 @@ propagate = false
 handlers = ["console"]
 
 [log_config.root]
-level = "WARNING"
+level = "DEBUG"
```
cverna: Was this only for debug ? or is it an intended change ?

Zlopez: This is config used for local testing in Vagrant only, so this is intended change.

PR #258, opened 2019-09-12
Coverage changes

Zlopez:
Lower coverage threshold to 50% to make tests pass, but add diff
coverage rule for 100%. This means that every new code must be covered
by tests.


pypingou:
:+1: thanks :)

PR #262, opened 2019-09-24
Increase test coverage

dcermak:
I have added some unit tests for the `buildsys.Koji` module: they mostly rely on mocking function calls and I am not 100% happy about them currently (as they are not really "deep") but I don't know a better way how to add tests without running a fully fledged integration test suite.

### DIFF
--- a/hotness/buildsys.py
+++ b/hotness/buildsys.py
@@ -31,6 +31,7 @@ import subprocess as sp
 import tempfile
 import threading
 import time
+import typing
 import copy
 
 from six.moves.urllib.parse import urlparse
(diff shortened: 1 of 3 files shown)


Zlopez:
I looked at the bodhi buildsys module and they actually have mock function for the API functions they are calling in Koji. So the Koji is one of the hard ones to test.

Zlopez:
Thank you for increasing the coverage, could you just squash the commits?

dcermak:
> Thank you for increasing the coverage, could you just squash the commits?

Yes, done.

PR #267, opened 2019-10-15
Increase test coverage of helpers.py

nielsenb-jf:
Modest increase to the test coverage of `helpers.py` (and shameless Hacktoberfest contribution). The resulting tests in turn produced the following changes:

- `rpm_max` and `upstream_max` have been updated to use sort `key` keyword argument instead of the non-existent in Python 3 `cmp` keyword argument, see e4bf65f.
- An unreachable section of `split_rc` was removed, see 77dcf15.
- `upstream_cmp` was changed to match the behavior described in a comments, and tested with the additional tests. See 19f70fe and 2e2c89b. This should be double-checked for correctness.

### DIFF
--- a/hotness/helpers.py
+++ b/hotness/helpers.py
@@ -18,6 +18,7 @@
 """
 
 import fnmatch
+import functools
 import re
 import pprint as pprint_module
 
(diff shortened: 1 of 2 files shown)


Zlopez:
Could you squash commits? Otherwise this looks good and I'm grateful for your PR.

nielsenb-jf:
I squashed it. Let me know if there are any other changes you want to see!

PR #271, opened 2020-01-17
Add expand_subdirs test

nielsenb-jf:
`url_prefix` check for empty string was removed as the code was unreachable. Since the `glob_pattern` has a minimum of two required `/` characters, there was no way for an url with less than 2 characters to match `glob_pattern`, so the `url_prefix` could never be an empty string.

### DIFF
--- a/hotness/tests/test_helpers.py
+++ b/hotness/tests/test_helpers.py
@@ -6,10 +6,73 @@ from __future__ import unicode_literals, absolute_import
 import re
 import unittest
 
+from unittest.mock import patch
 from hotness import helpers
 
 
 class TestHelpers(unittest.TestCase):
+    def test_expand_subdirs_no_match(self):
+        # Test no glob match
+        self.assertEqual(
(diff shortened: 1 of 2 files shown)


hotness/tests/test_helpers.py
```diff
@@ -6,10 +6,70 @@
 import re
 import unittest
 
+from unittest.mock import patch
 from hotness import helpers
 
 
 class TestHelpers(unittest.TestCase):
+    def test_expand_subdirs(self):
```
Zlopez: Could you split this into multiple tests? It's good to have one test case = one test.

nielsenb-jf: Sure thing. How about now?

Zlopez:
Everything looks fine, I just don't know why CI doesn't catch this PR.

Zlopez:
@nielsenb-jf Could you rebase this PR to latest master. I was able to finally solve the issue with CentOS CI.

nielsenb-jf:
Force pushed a rebased version.

PR #275, opened 2020-03-06
Fixes and tests for helpers.secure_download

nielsenb-jf:
StringIO has been replaced with io.StringIO, the former no longer exists in Python3.

This adds pycurl as an explicit requirement. pycurl is now required to be available in the tox environment, which in turn requires both libcurl-devel and openssl-devel installed in the vagrant environment. pycurl is told to use the correct ssl library by setting the PYCURL_SSL_LIBRARY in tox.ini.

### DIFF
--- a/devel/ansible/roles/hotness-dev/tasks/main.yml
+++ b/devel/ansible/roles/hotness-dev/tasks/main.yml
@@ -4,6 +4,8 @@
   dnf:
     name: [
       git,
+      libcurl-devel,
+      openssl-devel,
       ngrep,
       nmap-ncat,
       python3-rpdb,
--- a/hotness/helpers.py
+++ b/hotness/helpers.py
@@ -325,7 +325,7 @@ def filter_dict(d, key_list):
 
 def secure_download(url, cainfo=""):
     import pycurl
-    import StringIO
+    import io
 
     c = pycurl.Curl()
     c.setopt(pycurl.URL, url.encode("ascii"))
@@ -343,7 +343,7 @@ def secure_download(url, cainfo=""):
     if cainfo:
         c.setopt(pycurl.CAINFO, cainfo)
 
-    res = StringIO.StringIO()
+    res = io.StringIO()
 
     c.setopt(pycurl.WRITEFUNCTION, res.write)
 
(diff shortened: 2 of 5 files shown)


Zlopez:
The pycurl dependency is failing in CI with `FileNotFoundError: [Errno 2] No such file or directory: 'curl-config': 'curl-config'`.

@nielsenb-jf Could you look at it?

nielsenb-jf:
Did the vagrant environment get regenerated to pick up the updated main ansible roll?

`curl-config` should come along with the added `libcurl-devel` dependency. Also note `openssl-devel` is now required.

Zlopez:
The CI doesn't use the vagrant environment, it's just clean fedora with tox.

It looks like I need to create my own container for this and don't use the generic tox one, if we need some special devel packages.

nielsenb-jf:
That explains why I could not reproduce the behavior locally; I didn't try a run in plain tox. The Jenkins configuration does not appear to reside with the project so I can't really say much more.

Zlopez:
I created an issue for the changes, but I don't know when this will be ready. I will let you know, when the change in CI will be implemented.

Zlopez:
@nielsenb-jf The CI should be fixed now. Could you do a force push to restart the CI?

nielsenb-jf:
Pushed.

Zlopez:
The remaining errors are strange and I'm not sure if they are related to your PR in any way.

nielsenb-jf:
3 completely different errors, neat.

The error for f30 looks like it may be 'legitimate' in some form, with f30 going EOL in the not too distant future, I'm not sure how much digging into it I'm going to do. I'll at least look to see if I can reproduce it.

The f31 and f32 errors look like the cached psutil and bs4 wheels may have been bad somehow?

nielsenb-jf:
Cannot reproduce with plain tox on f30.

Zlopez:
@nielsenb-jf Then it's probably a temporary issue. I'm trying to now solve the ownership issue with the-new-hotness in CI and I will try to rerun the tests once it's solved. We will see if the issue is really just temporary.

PR #278, opened 2020-03-16
Add custom pipeline for CentOS CI

### LINKED ISSUE #277: Configure jenkins to use custom node instead of standard tox node
Because of the changes introduced in #275, we need to use custom node instead of the standard tox container.

This will need change in jenkins pipeline in CentOS CI. Maybe we could even use the [ansible provisioning script for vagrant](https://github.com/fedora-infra/the-new-hotness/tree/master/devel/ansible) to create the machine.

As a inspiration we can look at the [Bodhi implementation of Jenkins pipeline](https://github.com/fedora-infra/bodhi/tree/develop/devel/ci).

Zlopez:
Closes #277 

### DIFF
--- a/.cico.pipeline
+++ b/.cico.pipeline
@@ -1,1 +1,53 @@
-fedoraInfraTox { }
+#!groovy
+
+def call(Closure body) {
+    def stages = [:]
+    def fedora_containers = [
+                    containerTemplate(name: 'jnlp',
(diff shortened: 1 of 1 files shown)


Zlopez:
This PR is mistake, we will instead add the packages to fedora-tox containers in CentOS CI.

PR #285, opened 2020-04-23
Refactor config to use centralized defaults

### LINKED ISSUE #186: Refactor config to use centralized defaults.
See discussion in #185.

Zlopez:
Closes #186

### DIFF
--- a/hotness/config.py
+++ b/hotness/config.py
@@ -1,0 +1,172 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright © 2020  Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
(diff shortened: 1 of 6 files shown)


hotness/config.py
```diff
@@ -0,0 +1,172 @@
...
+        Merged configuration dictionary.
+    """
+    config = DEFAULTS.copy()
+
+    consumer_config = config_dict["consumer_config"]
+
+    _log.info("Loading the-new-hotness configuration")
+    for key in consumer_config:
+        try:
+            if config[key.upper()] is dict:
```
AdamSaleh, line 147: Wait, this is so that we can write config keys case-insensitive?

Zlopez: Yes, it will allow you to write config keys case-insensitive.

PR #289, opened 2020-05-04
Use anitya schema for fedora messaging

### LINKED ISSUE #248: Use Anitya schema for fedora-messaging
Currently the-new-hotness doesn't use Anitya fedora-messaging schema, but instead only expects some default message.

We should use Anitya schema to prevent any issue, when the schema changes.

Zlopez:
Closes #248

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,7 @@ Development
 * Remove unused functions (#280)
 * Clean and raise coverage for Anitya module (#281)
 * Change links to update policy (#291)
+* Use Anitya schema for fedora-messaging (#248)
 
 Features
 ^^^^^^^^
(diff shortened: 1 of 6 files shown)


Zlopez:
This PR will fail till the new version of [anitya_schema](https://pypi.org/project/anitya-schema/) will be released. I'm currently working on it.

Zlopez:
Ready to merge

Zlopez:
Right now, the message schema doesn't have any backwards incompatible changes. And if new version comes out, I will be notified by dependabot. Also I'm the one who is releasing the new version of Anitya schemas :-)

You can also find a [guide](https://fedora-messaging.readthedocs.io/en/stable/messages.html#message-conventions) in fedora-messaging how to introduce new versions of schemas.

PR #290, opened 2020-05-11
Fix crash on empty config section

Zlopez:
This fix is rather complex, so I split it to smaller parts:
1) Add recursive load of config dictionaries, this will ensure that
   every config option we know will be available and we don't end up
   with crash on empty dictionary
2) As direct reaction to the above default case sensitivity was changed
   to lower, because this caused issues when passing arguments as dictionary to
   3rd party libs
3) The copy of the config was changed to deep copy to also copy internal
   dictionaries

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,7 @@ Bugfixes
 ^^^^^^^^
 
 * Use fallback rpm comparison when rpm module is unavailable (#273)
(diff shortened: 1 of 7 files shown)


Zlopez:
The Dynaconf looks nice, but I started from what was already in Anitya and there is other reason I have this implementation. The config is not read from a file, but the file is already read by fedora-messaging library and Im just working with the dictionaries it passed to me.

PR #291, opened 2020-05-12
Change links to update policy

blowry-fedora:
Change the links to the update policy to point to the fedora docs site instead of the wiki.


blowry-fedora:
Ok, I've updated the changelog.

Zlopez:
Thanks

PR #304, opened 2020-07-14
Issue 302

### LINKED ISSUE #302: Errors while downloading sources
Hi,
I would like to ask if there is a way how to make automatic scratch builds
work if you have sources that are not located at some url but are directly in the
dist git. For example here https://bugzilla.redhat.com/show_bug.cgi?id=1835164
scratch build was not executed because the-now-hotness could not
download memcached.sysconfig which is a file stored directly in dist git and thus
does not need to be downloaded.

TomasKorbar:
Hi @Zlopez,
I decided to try to help you help me:)
So i fixed issue 302 with the commit https://github.com/fedora-infra/the-new-hotness/commit/03b2f6f9b48402fb1ae1b428a71a0398753391bd .
For more information please see commit messages.
Feel free to propose any change.

Closes #302

### DIFF
--- a/news/302.bug
+++ b/news/302.bug
@@ -1,0 +1,1 @@
+Errors while downloading sources
--- a/hotness/buildsys.py
+++ b/hotness/buildsys.py
@@ -34,7 +34,6 @@ import time
 import typing
 import copy
 
-from six.moves.urllib.parse import urlparse
 import koji
 
 from hotness import exceptions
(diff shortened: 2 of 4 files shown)


Zlopez:
I noticed that your PR is missing news file, could you add one.
You can look at how to add it in [docs](https://the-new-hotness.readthedocs.io/en/latest/dev-guide.html#release-notes).

TomasKorbar:
> I noticed that your PR is missing news file, could you add one.
> You can look at how to add it in [docs](https://the-new-hotness.readthedocs.io/en/latest/dev-guide.html#release-notes).

Fixed

news/302.bug
```diff
@@ -0,0 +1,5 @@
+Errors while downloading sources
```
Zlopez, line 1: It is better if the news file has only one line. This will be used to generate changelog.

TomasKorbar: Fixed.

PR #334, opened 2021-06-16
Remove Freenode, since we moved to libera

mscherer:


README.rst
```diff
@@ -15,7 +15,7 @@ To see recent messages from the-new-hotness:
 * Check Fedora's `datagrepper
   <https://apps.fedoraproject.org/datagrepper/raw?category=hotness&delta=2592000>`_
 
-* Or join #fedora-fedmsg IRC channel on freenode and watch for ``hotness``
+* Or join #fedora-fedmsg IRC channel on libera and watch for ``hotness``
```
Zlopez, line 18: It will be probably better to use libera.chat, because libera is showing multiple unrelated search result. And even better would be to have a link to https://libera.chat/

Zlopez (commented):
I'm not against the current state, but there is some space for improvement

Zlopez (approved):

Zlopez:
Thanks for the contribution :-)

PR #337, opened 2021-07-21
Containerized workflow

### LINKED ISSUE #336: Containerized workflow
# Description
Because the-new-hotness is running in OpenShift it will be nice to have similar environment for development. This means using containers instead of Vagrant.

This will be probably challenging because the-new-hotness is connecting to multiple external systems in Fedora infrastructure that need credentials, which are not provided in advance and you need to fill them before starting the-new-hotness service in vagrant.

## Requirements
* Create a container for the-new-hotness
* Create a container for RabbitMQ
* the-new-hotness container should be able to connect to RabbitMQ
* Update contribution guide
* (Optional) Use pytest-docker for integration testing
* (Optional) Add redis container

## How this will make the-new-hotness better
This will make production env and dev env almost one to one and will allow us to reproduce issues we encounter in production much more easily.

ozkanonur:
- [x] Create a container for the-new-hotness
- [x] Create a container for RabbitMQ
- [x] the-new-hotness container should be able to connect to RabbitMQ
- [x] Update contribution guide
- [ ] ~~(Optional) Use pytest-docker for integration testing~~
- [x] (Optional) Add redis container

fixes #336


ozkanonur:
I am not sure if we need to use pytest-docker for testing integration between services. I already check their healths (if they are available or not) before starting hostness service. What do you think? @Zlopez

Zlopez:
The healtcheck in docker-compose should be enough

ozkanonur:
You can review it when you have time, I think it's all done. I struggled writing the doc because I didn't write on Rst before. Sorry for pushing so many commits on that. You can squash the commits on merging the PR. @Zlopez

Dockerfile.dev
```diff
@@ -0,0 +1,33 @@
...
+
+# Package installation steps are separated so they
+# can be cached individually on Dockerfile changes
+
+RUN dnf upgrade --refresh -y
+
+# Kerberos packages
+RUN dnf install -y krb5-workstation krb5-devel krb5-libs
+
+# fedmsg (Federated Message Bus)
```
Zlopez, line 14: You can remove this, the fedmsg should be removed from the-new-hotness in favor of fedora-messaging

docker-compose.yml
```diff
@@ -0,0 +1,47 @@
...
+    healthcheck:
+      test: [ "CMD", "nc", "-z", "localhost", "5672" ]
+      interval: 3s
+      timeout: 3s
+      retries: 30
+    environment:
+      - RABBITMQ_DEFAULT_USER=hotness
+      - RABBITMQ_DEFAULT_PASS=hotness
+
+  redis:
```
Zlopez, line 37: The redis is not used yet, but it will be nice to have this container available prior to development.

docs/dev-guide.rst
```diff
@@ -74,6 +74,55 @@ You can also use aliases:
...
+* Docker
+* Docker Compose
+
+Next, clone the repository and start containers::
+
+    $ git clone https://github.com/fedora-infra/the-new-hotness.git
+    $ cd the-new-hotness
+    $ make up
+
+Hotness container starts after when Redis and RabbitMQ services successfully start. Usually, it takes around 10-30 seconds depends on the computer power.
```
Zlopez, line 93:
```suggestion
Hotness container starts after Redis and RabbitMQ services successfully start. Usually, it takes around 10-30 seconds depends on the computer power.
```

Zlopez (changes requested):
This looks good, however there are few things, that should be changed.

1. You should look at ansible roles in https://github.com/fedora-infra/the-new-hotness/tree/master/devel, documentation is outdated and this is how the-new-hotness is currently being provisioned in vagrant
2. Could we use podman instead of docker, it has compose support now

ozkanonur:
I renamed the build file with "Containerfile.dev" since we will also use podman. For using podman, we dont need to provide anything in compose or Containerfile, according to [this article](https://www.redhat.com/sysadmin/podman-docker-compose) only thing you need is `podman-docker`

docker-compose.yml
ozkanonur: Then, we can remove the condition of redis from hotness container for now.

ozkanonur:
I tried to implement things from Vagrant workflow, but as you already know we can not do things like starting system services, using systemctl, using motd(I did some dirty workflow to provide the same functionality, you can check [here](https://github.com/ozkanonur/the-new-hotness/blob/master/Makefile#L8)). So I am not sure about [this bashrc](https://github.com/ozkanonur/the-new-hotness/blob/master/devel/ansible/roles/hotness-dev/files/bashrc). Does it have to be run as a system service?

Zlopez:
> I tried to implement things from Vagrant workflow, but as you already know we can not do things like starting system services, using systemctl, using motd(I did some dirty workflow to provide the same functionality, you can check [here](https://github.com/ozkanonur/the-new-hotness/blob/master/Makefile#L8)). So I am not sure about [this bashrc](https://github.com/ozkanonur/the-new-hotness/blob/master/devel/ansible/roles/hotness-dev/files/bashrc). Does it have to be run as a system service?

Don't bother with systemd services, just use the ansible as help when setting up the containers. It would be nice if the functionality inside the container will remain same, but if you do it by other means, it doesn't matter. Just document it.

ozkanonur:
> > I tried to implement things from Vagrant workflow, but as you already know we can not do things like starting system services, using systemctl, using motd(I did some dirty workflow to provide the same functionality, you can check [here](https://github.com/ozkanonur/the-new-hotness/blob/master/Makefile#L8)). So I am not sure about [this bashrc](https://github.com/ozkanonur/the-new-hotness/blob/master/devel/ansible/roles/hotness-dev/files/bashrc). Does it have to be run as a system service?
> 
> Don't bother with systemd services, just use the ansible as help when setting up the containers. It would be nice if the functionality inside the container will remain same, but if you do it by other means, it doesn't matter. Just document it.

Alright then, I will update motd and docs now. Also, what do you think about removing tmux, tree, ngrep, nmap-ncat ? Using such tools like that is a bit far from container philosophy. What I recommend is let's keep that tools outside of the containers and not use them in containers.

Zlopez:
I don't think we need these tools in container, so they can be safely removed. If somebody needs them they can install them on demand inside container itself.

ozkanonur:
Can you tell me if anything missing on PR right now?

docs/dev-guide.rst
```diff
@@ -74,6 +74,55 @@ You can also use aliases:
...
+* ``make restart`` Restarts all the container services that are either stopped or running 
+* ``make halt`` Stops and removes the containers
+* ``make bash`` Connects to hotness container
+* ``make logs`` Shows all logs of all containers
+
+Project files are bound to each other with host and container. Whenever you change any project file from the host or the container, the same change will happen on the opposite side as well.
+
+After connecting to hotness container you can run the applicaton with::
+
+    $ fedmsg-hub
```
Zlopez, line 124: This is outdated way of running the-new-hotness. The new way is running it as described in [this systemd service](https://github.com/fedora-infra/the-new-hotness/blob/master/devel/ansible/roles/hotness-dev/files/hotness.service). You need to set the `FEDORA_MESSAGING_CONF` variable and run `fedora-messaging consume`

Zlopez (changes requested):
It's starting to look good, just few more things I noticed.

Could you also add news file for towncrier, see [contribution guidelines](https://the-new-hotness.readthedocs.io/en/stable/dev-guide.html#release-notes) for more info.

Containerfile.dev
```diff
@@ -0,0 +1,40 @@
...
+    python3-fedmsg python3-koji python3-pycurl python3-requests python3-six \
+    python3-black python3-flake8 python3-sphinx
+
+# Development tools (for doing stuff inside of the container)
+RUN dnf install -y vim-enhanced tox python3-rpdb
+
+# Packages that are required for releasing
+RUN pip install wheel twine towncrier
+
+COPY . .
```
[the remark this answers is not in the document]
ozkanonur: We need to copy the whole folder to the inside of the container for the following steps. But, we are also mounting the same folder to the inside of the container and that overrides everything back. So if you do changes inside of the container, you will have them outside of the container as well. Only if you do changes at building steps then you won't see the changes outside. But we are not doing it so.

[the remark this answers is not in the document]
Zlopez: You are right, I didn't noticed the volumes in compose file.

ozkanonur:
> It's starting to look good, just few more things I noticed.
> 
> Could you also add news file for towncrier, see [contribution guidelines](https://the-new-hotness.readthedocs.io/en/stable/dev-guide.html#release-notes) for more info.

News folder is already provided in the container, I couldn't see any other related files that I should provide. Am I missing something?

ozkanonur:
Also `amqp_url` updated to ``amqp://hotness:hotness@localhost:5672`` from ``amqp://``.

Zlopez:
> News folder is already provided in the container, I couldn't see any other related files that I should provide. Am I missing something?

I meant that you should add a news file for this PR, so is't in the changelog


> Also `amqp_url` updated to `amqp://hotness:hotness@localhost:5672` from `amqp://`. But I am not sure if we should do this. Because not everybody will do development in containers. Tell me if we should keep or remove it back. @Zlopez

If this will work with container deployment I'm in for keeping it, or we can create separate config file for container setup.

ozkanonur:
> > News folder is already provided in the container, I couldn't see any other related files that I should provide. Am I missing something?
> 
> I meant that you should add a news file for this PR, so is't in the changelog
> 
> > Also `amqp_url` updated to `amqp://hotness:hotness@localhost:5672` from `amqp://`. But I am not sure if we should do this. Because not everybody will do development in containers. Tell me if we should keep or remove it back. @Zlopez
> 
> If this will work with container deployment I'm in for keeping it, or we can create separate config file for container setup.

I see, created the news file now.

Lets keep it then, you should be able to connect to RabbitMQ by using that url without any issue.

Makefile
```diff
@@ -0,0 +1,11 @@
+up:
+	docker-compose up -d hotness
```
Zlopez, line 2: Will this work on podman only system?

news/_template.rst
```diff
@@ -4,7 +4,7 @@
    {%- elif value.startswith("C") -%}
      `{{ value[1:] }} <https://github.com/fedora-infra/the-new-hotness/commit/{{ value[1:] }}>`_
    {%- else -%}
-     `#{{ value }} <https://github.com/fedora-infra/the-mew-hotness/issues/{{ value }}>`_
+     `#{{ value }} <https://github.com/fedora-infra/the-new-hotness/issues/{{ value }}>`_
```
Zlopez, line 7: Thanks for catching this out, I didn't noticed it.

docs/dev-guide.rst
```diff
@@ -74,6 +74,55 @@ You can also use aliases:
...
+* Docker / Podman (version +3 with podman-docker)
+* Docker Compose
+
+Next, clone the repository and start containers::
+
+    $ git clone https://github.com/fedora-infra/the-new-hotness.git
+    $ cd the-new-hotness
+    $ make up
+
+Hotness container starts after it's depended container successfully start. Usually, it takes around 10-30 seconds depends on the computer power.
```
Zlopez, line 93:
```suggestion
Hotness container starts after start start of containers running services required by the-new-hotness. Usually, it takes around 10-30 seconds depends on the computer power.
```

Zlopez (changes requested):
Just one more thing and one question. Otherwise :+1:

news/336.feature
```diff
@@ -0,0 +1 @@
+Provide Containerized workflow
```
Zlopez, line 1: This should be a `.dev` change, not feature.

ozkanonur: I was confused between those two. Just as a suggestion, if we label the issues with those things and then update the documentation with `filename should be "{issue-id}.{issue-label}"`. I think this can be much cleaner to do.

Makefile
ozkanonur: Yes, but it requires version +3.

Articles about that:
- https://podman.io/blogs/2021/01/11/podman-compose.html
- https://www.redhat.com/sysadmin/podman-docker-compose

news/336.feature
Zlopez: Good idea, I will create a ticket for it.

Makefile
Zlopez: I know about the version restriction, although I wasn't sure, if you can use `docker-compose` instead of `podman-compose`.

ozkanonur:
Since we will use this for also podman, I renamed compose to container-compose. I think its better to not call anything named `docker`. Only the makefile includes that keyword now.

news/336.feature
Zlopez: https://github.com/fedora-infra/the-new-hotness/issues/338

Zlopez:
Everything looks fine now, could you just squash the commits before merging?

ozkanonur:
> Everything looks fine now, could you just squash the commits before merging?

I dont have access to my computer right now. Here should be option like `squash and merge`. If it doesnt exists, I can do it today around 17.00 UTC

Zlopez:
Thanks for your work, now I will need to start using the container workflow instead vagrant :-D

ozkanonur:
> Thanks for your work, now I will need to start using the container workflow instead vagrant :-D

Thank you too for letting me be part of this. If anything is not stable tell me via mail that you can find on my profile.

PR #339, opened 2021-07-23
Sync issues with news

### LINKED ISSUE #338: Synchronize GitHub issue labels with news file types
It would be easier to comprehend if there were labels connected to news file types. The documentation will then contain this description of news file `{issue-id}.{issue-label}`, which will be much more easier to understand than the current description.

ozkanonur:
Fixes #338

### DIFF
--- a/docs/dev-guide.rst
+++ b/docs/dev-guide.rst
@@ -227,7 +227,7 @@ News type can be one of the following:
   
 For example:
 
-If this PR is solving issue #714 with labeled as ``type.bug``, the file inside news should be called 714.bug (PR714.bug if the PR does not related with any issue) and the content of the file would be:
+If this PR is solving issue #714 labeled as ``type.bug`` and named "Javascript error on add project page", the file inside news should be called 714.bug (PR714.bug if the PR does not have any linked issue and the PR number is 714) and the content of the file would be:
 
 ``Javascript error on add project page``
 
@@ -323,4 +323,3 @@ If you are a maintainer and wish to make a release, follow these steps:
 
 .. note::
     Optional steps are required only if you want to release a new version of message schema.
-
--- a/docs/dev-guide.rst
+++ b/docs/dev-guide.rst
@@ -206,16 +206,28 @@ Release notes
 =============
 
 To add entries to the release notes, create a file in the ``news`` directory
-with the ``id.type`` name format, where ``type`` is either label of the
-issue or the pull request.
+with the ``source.type`` name format, where the ``source`` part of the filename is:
 
-And where the ``id`` part of the filename is either id of the issue or the pull
-request.
+* ``42`` when the change is described in issue ``42``
+* ``PR42`` when the change has been implemented in pull request ``42``, and
+  there is no associated issue
+* ``username`` for contributors (``author`` extention). It should be the
+  username part of their commit's email address.
 
+And where ``type`` is label of the issue or PR that is named ``type.label``. If the issue or PR is missing a label, please ask maintainer to add one.
+
+News type can be one of the following:
+
+* ``feature``: for new features
+* ``bug``: for bug fixes
+* ``api``: for API changes
+* ``dev``: for development-related changes
+* ``author``: for contributor names
+* ``other``: for other changes
+  
 For example:
 
-If this PR is solving issue #714 with labeled as ``type:bug``, the file inside ``news`` should be called ``714.bug`` (``PR714.bug`` if the PR does not related with any issue)
-and the content of the file would be:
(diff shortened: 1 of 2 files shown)


ozkanonur:
I tried to simplify docs but for merging this, we need to provide new labels (``type:bug``, ``type:feature``, etc) and attach them to current issues (or pull requests that are not related to issues)

Zlopez:
Something must be wrong, I see 40 commits in this PR. Could you rebase it to latest master?

ozkanonur:
> Something must be wrong, I see 40 commits in this PR. Could you rebase it to latest master?

Sorry, I didnt notice that. Should be fine now.

docs/dev-guide.rst
```diff
@@ -206,26 +206,15 @@ Release notes
 =============
 
 To add entries to the release notes, create a file in the ``news`` directory
-with the ``source.type`` name format, where ``type`` is one of:
-
-* ``feature``: for new features
```
Zlopez, line 211: I don't think it's good idea to remove this text. We should still have list of supported labels.

docs/dev-guide.rst
```diff
@@ -206,26 +206,15 @@ Release notes
...
-* ``bug``: for bug fixes
-* ``api``: for API changes
-* ``dev``: for development-related changes
-* ``author``: for contributor names
-* ``other``: for other changes
-
-And where the ``source`` part of the filename is:
```
Zlopez, line 218: Same as before, this still needs to be described here.

docs/dev-guide.rst
```diff
@@ -206,26 +206,15 @@ Release notes
...
-
-* ``42`` when the change is described in issue ``42``
-* ``PR42`` when the change has been implemented in pull request ``42``, and
-  there is no associated issue
-* ``username`` for contributors (``author`` extention). It should be the
-  username part of their commit's email address.
-  
+with the ``id.type`` name format, where ``type`` is either label of the
```
Zlopez, line 209: It's not clear from this descriptiong what label should be used. I would use something like `type is label of the issue that is named label.type (see supported types above)`. If the issue or PR is missing a label, please ask maintainer to add one.

docs/dev-guide.rst
```diff
@@ -206,26 +206,15 @@ Release notes
...
+issue or the pull request.
+
+And where the ``id`` part of the filename is either id of the issue or the pull
```
Zlopez, line 212: This is described clearly above.

news/338.dev
```diff
@@ -0,0 +1 @@
+Update filename format of news to sync issues with news
```
Zlopez, line 1:
```suggestion
Update filename format of news file to correspond with labels for issues
```

Zlopez (changes requested):
I don't think we should remove the whole section that was there before, this will still be true, we just add the label so it's more clear to know what type to use.

One more thing, could you switch `source` and `type` explanation in the text, so it's in the same order as in the name of the file. `source` explanation first, `type` second.

docs/dev-guide.rst
ozkanonur: I thought `label` points to the `type`. Shouldn't be something like `id.label` instead of `label.type`?

Zlopez: I wrote it backwards, it should be the name of the label `type.label`, so it's easy to recognize among other labels.

docs/dev-guide.rst
```diff
@@ -206,27 +206,28 @@ Release notes
...
-* ``PR42`` when the change has been implemented in pull request ``42``, and
-  there is no associated issue
-* ``username`` for contributors (``author`` extention). It should be the
-  username part of their commit's email address.
   
 For example:
 
-If this PR is solving bug 714 the file inside ``news`` should be called ``714.bug``
-and the content of the file would be:
+If this PR is solving issue #714 with labeled as ``type.bug``, the file inside news should be called 714.bug (PR714.bug if the PR does not related with any issue) and the content of the file would be:
```
Zlopez, line 230:
```suggestion
If this PR is solving issue #714 labeled as ``type.bug`` and named "Javascript error on add project page", the file inside news should be called 714.bug (PR714.bug if the PR does not have any linked issue and the PR number is 714) and the content of the file would be:
```

Zlopez (changes requested):
Just some wording.

ozkanonur:
> Just some wording.

Sorry, I shouldn't open this PR probably since my grammar is not perfect :D

Zlopez (approved):
Now I just need to create the labels :-D

Zlopez:
The labels are now available, merging this PR.

PR #340, opened 2021-07-24
Provide podman compatibility

ozkanonur:
I tested the container environment on RHEL 8 and faced with following errors:

- Somehow getting `pip not found` and `python not found`
- Healthchecks are not recognized by podman, so we need to either comment out `depends_on` block of `hotness` service or all the `healtcheck` blocks.

On this PR, I provide Containerfile compatibility for podman, and updated the docs for letting the contributors know that they should comment `depends_on` or `healtcheck` blocks for using podman.

### DIFF
--- a/Containerfile.dev
+++ b/Containerfile.dev
@@ -14,8 +14,8 @@ RUN dnf install -y krb5-devel
 # Fedora Messaging
 RUN dnf install -y fedora-messaging
 
-# Python packages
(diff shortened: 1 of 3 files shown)


Zlopez (approved):
This looks fine, I'm surprised that the pip error didn't show up earlier.

ozkanonur:
> This looks fine, I'm surprised that the pip error didn't show up earlier.

Same. Maybe they removed python from the base image?

PR #405, opened 2022-02-14
Remove python 3.7

Zlopez:
With introduction of Sphinx 4.4.0 the dependencies couldn't be satisfied on
Python 3.7, so it's probably time to remove it.

```
The conflict is caused by:
    flake8 4.0.1 depends on importlib-metadata<4.3; python_version < "3.8"
    pytest 6.2.5 depends on importlib-metadata>=0.12; python_version < "3.8"
    sphinx 4.4.0 depends on importlib-metadata>=4.4; python_version < "3.10"
```

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -52,10 +52,6 @@
             vars:
               dependencies:
                 - krb5-devel
-        - tox-python37:
-            vars:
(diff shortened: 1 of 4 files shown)


Zlopez:
recheck

PR #434, opened 2022-04-05
Use default tox job instead of using the custom one

Zlopez:
It would be much more convenient to reuse existing zuul tox job instead of
creating multiple simple jobs.

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -43,39 +43,61 @@
 - project:
     check:
       jobs:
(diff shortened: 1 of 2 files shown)


Zlopez:
We can't use the existing tox job for multiple checks, because the name needs to be different for every job, so if we want to run multiple tox jobs, we need multiple names for them :-/

But we can at least do some changes to make them more convenient.

PR #519, opened 2022-09-25
fix the moved explanation url

erolkskn:
It's a simple one line change to config. Old explanation url moved to new one. 
See: https://fedoraproject.org/wiki/Upstream_release_monitoring


Zlopez:
Could you add news file as well?

Zlopez:
And it would be nice to change it [here](https://pagure.io/fedora-infra/ansible/blob/main/f/roles/openshift-apps/the-new-hotness/templates/config.toml) as well. This is the configuration file that is used for deployment in both https://stg.release-monitoring.org and https://release-monitoring.org

erolkskn:
> And it would be nice to change it [here](https://pagure.io/fedora-infra/ansible/blob/main/f/roles/openshift-apps/the-new-hotness/templates/config.toml) as well. This is the configuration file that is used for deployment in both https://stg.release-monitoring.org and https://release-monitoring.org

Thanks for pointing it out Michal :-) I'm going to take a look at it now

Issue #534, opened 2022-10-12
bug reported for "latest" version instead of newly "retrieved" version


decathorpe:
originally reported here:
https://github.com/fedora-infra/anitya/issues/1496

It looks like the-new-hotness uses either the "biggest" version when filing new bugs on RHBZ instead of the one that was actually retrieved.

This can result in bugs getting reported for the wrong version, for example, if a version filter was set up to exclude pre-releases but there already were pre-releases in anitya's database.

Zlopez:
Fixed in https://github.com/fedora-infra/the-new-hotness/pull/537

PR #537, opened 2022-11-22
Fix the package version reported in the bugzilla header

LenkaSeg:
Related issue #534

### DIFF
--- a/hotness/hotness_consumer.py
+++ b/hotness/hotness_consumer.py
@@ -73,6 +73,8 @@ class HotnessConsumer(object):
 
     Attributes:
         short_desc_template (str): Short description template for notifier
+        short_desc_template_more_versions (str): Short description template for notifier
+                                                for more versions
         description_template (str): Template for the message content
         dist_git_url (str): URL for dist-git server
         distro (str): Distro to watch the updates for
@@ -117,6 +119,9 @@ class HotnessConsumer(object):
 
         # Initialize attributes
         self.short_desc_template = config["bugzilla"]["short_desc_template"]
+        self.short_desc_template_more_versions = config["bugzilla"][
+            "short_desc_template_more_versions"
+        ]
         self.description_template = config["bugzilla"]["description_template"]
         self.dist_git_url = config["dist_git_url"]
         self.explanation_url = config["bugzilla"]["explanation_url"]
@@ -601,7 +606,7 @@ class HotnessConsumer(object):
         bz_id = -1
         latest_upstream = package.version
 
-        upstream_versions = latest_upstream
+        upstream_versions = package.version
         if retrieved_versions:
             upstream_versions = ", ".join(retrieved_versions)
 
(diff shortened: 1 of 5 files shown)


LenkaSeg:
Still working on tests.

hotness/hotness_consumer.py
```diff
@@ -619,12 +619,15 @@ def _comment_on_bugzilla_with_template(
...
+        versions = package.version if package.version in retrieved_versions else retrieved_versions
+        if len(versions) > 1:
+            versions = f"New versions of {package.name} available."
         notify_request = NotifyRequest(
             package=package,
             message=description,
             opts={
                 "bz_short_desc": self.short_desc_template
-                % dict(name=package.name, latest_upstream=package.version)
+                % dict(name=package.name, latest_upstream=versions)
```
Zlopez, line 630: This will not work, because this is adding to template. It will either produce something like this "<package.name>-New versions of <package.name> available is available" for multiple versions.
You will need to instead replace `bz_short_desc` key in `opts` dict.

LenkaSeg: I see on the tests, it was only first shot :)

hotness/hotness_consumer.py
```diff
@@ -601,16 +606,14 @@ def _comment_on_bugzilla_with_template(
...
-        if retrieved_versions:
-            upstream_versions = ", ".join(retrieved_versions)
-
         # Prepare dist-git URL
         dist_git_url = self.dist_git_url + "/rpms/" + package.name
         # Prepare message for bugzilla
         description = self.description_template % dict(
             latest_upstream=package.version,
-            retrieved_versions=upstream_versions,
+            retrieved_versions=", ".join(retrieved_versions)
```
Zlopez, line 614: Not easily readable, it will be better to fill this variable before calling the `dict` function.

Zlopez: The same was available on the line 604 you deleted :-)

tests/test_config.py
```diff
@@ -53,6 +53,7 @@
             "reporter": "",
             "reporter_email": "",
             "short_desc_template": "",
+            "short_desc_template_more_versions": "New versions of %(name)s available.",
```
Zlopez, line 56: This could be empty, the test is just checking if it's available.

tests/test_hotness_consumer.py
```diff
@@ -410,6 +411,146 @@ def test_call_anitya_update_all_versions(self):
             package, "update.bug.file", exp_opts
         )
 
+    def test_bugzilla_notify_with_retrieved_versions(self):
```
Zlopez, line 414: I know this is here in other tests, but you shouldn't run more than one test case in one method. Either split this up or use [pytest parametrize](https://docs.pytest.org/en/6.2.x/parametrize.html).

Zlopez (changes requested):
Few minor issues

LenkaSeg:
pytest.parametrize makes the test much less voluminous, thanks for the suggestion!

Zlopez (approved):

Zlopez:
Before moving this to staging, we need to update staging config as well to reflect the new configuration values.

PR #559, opened 2023-04-21
Change whitelist to allowlist

Zlopez:
Tox changed the whitelist to allowlist and break the tests. Let's fix that and hope that this play with words doesn't happen in the future.


Zlopez:
It seems that the tests are failing due to some pip error. Let me merge it even with the errors.

PR #583, opened 2023-10-15
Drop query to PDC

LenkaSeg:
Remove PDC calls from the-new-hotness.

PDC was used when checking if a package is retired. Now it is checked on dist-git. 

Module `pdc.py` renamed to `retired.py`. I'm not sure if it's the best approach, since the other modules are named after the app they are calling, but naming it `dist-git` for example might be confusing, since there is a `pagure` module which also uses dist-git. 
Another option would be to add the validator of the retired packages as a method of the `pagure` class. That would need wider changes in the code I think, but it's doable :)


docs/ca-design.rst
```diff
@@ -253,9 +253,9 @@ defined in use cases layer.
     Class that is checking if the package is newer or not than the package currently available
     in Fedora using mdapi system. Inherits from `validator.py`.
 
-  * **pdc.py**
+  * **retired.py**
```
Zlopez, line 256: There should be one validator per external system (it makes it easier to replace them when needed). In this case it would be better to update the Pagure validator and add a new key `retired` to output dictionary.

Zlopez:
The implementation should be different and adhere to original design. One external system per one validator. Also the news file is missing.

docs/ca-design.rst
LenkaSeg: Added `retired` as a new key in the output dictionary under the `pagure` validator. 
But I'm not too happy about having two request calls in one function. It complicates the testing and does not feel right. I'll try to figure out how to make it better.

LenkaSeg:
Ah, yes, and the news file. Changing this PR to draft for now.

LenkaSeg:
Formatting fails on black failure to reformat a file with `match`.
Here related issue in black repo: https://github.com/psf/black/issues/2242
It has been fixed but `black` command need to be changed to `black --target-version py310`

LenkaSeg:
Mmm...python 38 and 39 are also not happy about case matching...

hotness/validators/pagure.py
```diff
@@ -56,7 +56,7 @@ class Pagure(Validator):
     about new version and optionally run a scratch build.
 
     Attributes:
-        url: URL of the PDC server
```
Zlopez, line 59: I didn't even knew I had this in the code.

hotness/validators/pagure.py
```diff
@@ -56,7 +56,7 @@ class Pagure(Validator):
...
+        url: URL of the dist-git server
```
Zlopez, line 59:
```suggestion
        url: URL of the pagure dist-git server
```

Zlopez (commented):
Only one minor thing

Zlopez (approved):

PR #611, opened 2025-07-21
Add rhbz prefix in changelog update message

dreua:
With this prefix, bodhi automatically links the bug to the update and closes the bug accordingly.

### DIFF
--- a/hotness/builders/koji.py
+++ b/hotness/builders/koji.py
@@ -132,7 +132,7 @@ class Koji(Builder):
 
             specfile = os.path.join(tmp, package.name + ".spec")
 
-            comment = "Update to %s (#%d)" % (package.version, bz_id)
+            comment = "Update to %s (rhbz#%d)" % (package.version, bz_id)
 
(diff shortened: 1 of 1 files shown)


dreua:
I'll wait for some feedback if that change is appreciated in general before I might consider deep diving into the testing framework to fix those tests.

Should we also update the [description here](https://github.com/fedora-infra/the-new-hotness/blob/4b7e08c8619eb687d07fa8803290d03a140042a9/hotness/patchers/bugzilla.py#L109) while were at it? (There seems to be a little code duplication going on here tbh.)

Zlopez:
That is a great improvement. If you will be able to fix the CI errors and add news file I will merge it.

> Should we also update the [description here](https://github.com/fedora-infra/the-new-hotness/blob/4b7e08c8619eb687d07fa8803290d03a140042a9/hotness/patchers/bugzilla.py#L109) while we're at it? (There seems to be a little code duplication going on here tbh.)

That one is to attach patch to bugzilla ticket, feel free to update it as well.

PR #626, opened 2025-12-20
Implement retry mechanism for transient failures

### LINKED ISSUE #322: Let's try to resubmit messages when error happens
Currently the-new-hotness is try only once to publish Fedora messages, it will be nice if we try it a few more times.

Prachi194agrawal:
This PR implements a retry mechanism for transient failures as requested in #322.

Previously, the-new-hotness would only attempt to process a message once. If a network timeout or temporary API outage occurred (e.g., Bugzilla or Koji being unreachable), the message would be lost.

Changes
hotness/hotness_consumer.py: Wrapped the message handling logic in a try-except block.

Transient network errors (requests.exceptions.RequestException) now raise fedora_messaging.exceptions.Nack, which triggers a re-queueing of the message for a later retry.

General exceptions (permanent failures) are logged, and the message is acknowledged (dropped) to prevent infinite loops.

hotness/use_cases/package_check_use_case.py: Updated to re-raise network exceptions instead of catching and returning a ResponseFailure, allowing the consumer to handle the retry logic.

Tests: Added 4 new test cases to tests/test_hotness_consumer.py to verify:

Retry on ConnectionError.

Retry on Timeout.

Retry on general RequestException.

Message drop on ValueError (permanent failure).

Fixes #322
Screenshot from 2025-12-20 16-42-14

### DIFF
--- a/hotness/hotness_consumer.py
+++ b/hotness/hotness_consumer.py
@@ -197,8 +197,8 @@ class HotnessConsumer(object):
                 self._handle_anitya_version_update(message)
             elif topic.endswith("buildsys.task.state.change"):
                 self._handle_buildsys_scratch(msg)
-        except requests.exceptions.RequestException as e:
-            # This catches Timeout, ConnectionError, and other transient network issues
+        except (requests.exceptions.ConnectionError, requests.exceptions.Timeout) as e:
+            # This catches Timeout and ConnectionError (transient network issues)
             _logger.warning(
                 "Transient network error processing message %s: %s. "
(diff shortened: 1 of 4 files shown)


Zlopez (commented):
The PR looks good, just few things:

* Missing news file. See https://the-new-hotness.readthedocs.io/en/stable/dev-guide.html#release-notes
* One question about the request exception as I couldn't find the information in requests documentation

Zlopez (approved):
I totally missed the update, now it looks good.

